PR #47, opened 2017-03-17
Drop the virtualenv dependency

brettcannon:
Venv as included in the Python 3.6 stdlib works just fine.


Mariatta:
Thanks :) Please merge it.

Issue #51, opened 2017-03-17
When cherry_picker fails, drop you off in the checkout


brettcannon:
I just had a cherry-pick fail, but because I was running `cherry_picker` from its own git repo it took me a minute to realize that I couldn't find the failed branch because I had not `cd cpython` yet. It might work out better to automatically `cd` into the checkout upon failure (with an appropriate message that this was done).

Mariatta:
Sorry about that.. yeah it's definitely something that I need to address. Maybe related to https://github.com/python/core-workflow/issues/45 too

brettcannon:
@Mariatta Nothing to apologize for! Tool worked out great (only issue is GitHub seems to keep thinking I might want to create a PR for the 3.5 branch)!

ncoghlan:
If we did decide to do this (and I think it's reasonable for the interactive case), it would be helpful if it printed two things before exiting:

* a command to finish just the "commit-and-push-and-open-web-browser" steps for the current target branch
* a command to run the cherry-pick again for any not yet processed branches

Mariatta:
Hmm.. I couldn't figure out how to actually ``cd`` to the ``cpython`` directory and let it persist there after cherry_picker.py ended 😅  So instead I'm printing out the message that you need to ``cd cpython`` first..

Mariatta:
Once https://github.com/python/core-workflow/pull/99 is merged, this is not needed anymore since the people will be cherry-picking from cpython directory.

Issue #45, opened 2017-03-16
Consider how to handle failed cherry-picks


Mariatta:
_From @ncoghlan on March 12, 2017 4:57_

I think this utility has a lot of potential to evolve into the bot that actually does the automated cherry-pick PR creation, and one of the things we'll need to address there is what to do when the routine cherry pick fails.

At the moment, if the first cherry-pick fails, it leaves the checkout in a messy state, so all the subsequent branch creation commands fail.

I'm thinking one way of tackling this would be to:

Complain and exit early if the checkout is in a messy state, and suggesting running git stash to move the local changes aside.
If any of the cherry-picks fail do an immediate git reset --hard and move on to the next branch.
Does that approach sound reasonable?

_Copied from original issue: Mariatta/chic_a_cherry_picker#15_

ncoghlan:
https://github.com/python/core-workflow/issues/51 points out that for the interactive use case, dropping the user into the failed cherry-pick directory with a printed command on:

* how to finish the current cherry-pick
* how to resume any other not-yet-attempted cherry-picks

serhiy-storchaka:
Would be nice if `cherry_picker.py` be allowing to fix errors manually and continue its work when run with `--continue`. Running with `--abort` should abort the current conflict resolution process, reset the workspace and delete just created branch.

Mariatta:
Thanks @serhiy-storchaka . I started working on handling errors in cherry_picker.py see #57 
I implemented the interactive mode where it will ask the user of how to proceed.
But if we want it as ``--continue`` or ``--abort`` options, it can be done.

ncoghlan:
The `--continue`/`--abort` approach would align nicely with the way git's own commands work, so +1 from me for that approach.

Mariatta:
Please check if https://github.com/python/core-workflow/pull/65 does what's expected. Thanks :)

PR #57, opened 2017-03-22
wip: Handle error when cherry-picking

Mariatta:
Attempt at handling errors while cherry-picking.

demo: https://asciinema.org/a/108298

When an error occurs, present two options to user:
1. Stop cherry-picking, and continue manually
2. Skip the current branch, and try to cherry-pick the next branch

When 1 is selected, provide hints on how to continue

When 2 is selected, abort the cherry-pick and continue with the next branch

### DIFF
--- a/cherry_picker/cherry_picker.py
+++ b/cherry_picker/cherry_picker.py
@@ -2,6 +2,7 @@ import click
 import os
 import subprocess
 import webbrowser
+import sys
 
 
 @click.command()
@@ -13,6 +14,7 @@ import webbrowser
 def cherry_pick(dry_run, pr_remote, commit_sha1, branches):
     if not os.path.exists('./pyconfig.h.in'):
         os.chdir('./cpython/')
+
     upstream = get_git_fetch_remote()
     username = get_forked_repo_name(pr_remote)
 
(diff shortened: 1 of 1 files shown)


Mariatta:
This PR attempts to address some of the comments in https://github.com/python/core-workflow/issues/51 and https://github.com/python/core-workflow/issues/45

In case there's an error in cherry-picking, it will enter an interactive mode, where user can decide whether they want to quit, or to continue with backport of the next branch.

Please see the demo of how this works at https://asciinema.org/a/108298
Looking for feedback about this approach.

Thanks :)

brettcannon:
I've only watched the video, but one tweak I would suggest making is when you say "aborting ..." I would name what Python version you're aborting for (e.g. "aborting for 2.7 ..." or something), otherwise I initially thought the whole process was aborting.

serhiy-storchaka:
I think the script always should stop cherry-picking and save the current state in local or global file if it can't be derived from the state of git. It is very unlikely that cherry-picking to 3.5 can be successful after failing cherry-picking to 3.6. The script should be able to skip already passed steps when run with the `--continue` option.

brettcannon:
So I guess that would require sorting the branches to make sure it goes from newest to oldest to make sure your failures fall as late as possible?

Mariatta:
I made #59 so it will now cherry-pick from the newest branch.

Mariatta:
Closing. Please see https://github.com/python/core-workflow/pull/65 instead.

PR #59, opened 2017-03-23
cherry-pick from the newest branch

Mariatta:


cherry_picker/cherry_picker.py
```diff
@@ -26,7 +26,7 @@ def cherry_pick(dry_run, pr_remote, commit_sha1, branches):
     if not branches:
         raise ValueError("at least one branch is required")
 
-    for branch in branches:
+    for branch in sorted(branches, reverse=True):
```
ncoghlan: This will already do the right thing for X.Y branches, but do we want to future-proof it against the day of 3.10 by adding `key=lambda v: tuple(map(int, v.split('.')))`?

Mariatta:
Thanks :) I updated the code anyway, and added a test for it.

brettcannon:
Maybe it's time to set up Travis if there's going to be tests?

Mariatta:
Sounds good :)

Issue #64, opened 2017-03-24
Make cherry_picker installable and create script


pitrou:
It cherry_picker were installable and created an executable script, it would make it easier to symlink it into PATH, rather than have to do the environment activation dance then `python -m cherry_picker` in the right directory.

PR #73, opened 2017-04-22
Make cherry picker as a script

lulouie:
Relate to #64 #69 

*I'm not sure if this change will break other bot or something.*

This commit make cherry-picker more easier to use, you don't need to let `cpython` folder under `cherry_picker` anymore.

Example:
```
$ ls
cpython   # Assume we already clone cpython in our dev dir and upstream setting
$ git clone https://github.com/python/core-workflow.git
$ cd core-workflow
$ python -m venv venv
$ source venv/bin/activate
(venv) $ python -m pip install -r cherry_picker/requirements.txt
(venv) $ cp -r cherry_picker venv/lib/python3.6/site-packages/
(venv) $ cd ..
(venv) $ python -m cherry_picker deadbeaf 3.6
```

### DIFF
--- a/cherry_picker/__main__.py
+++ b/cherry_picker/__main__.py
@@ -1,0 +1,4 @@
+from .cherry_picker import cherry_pick_cli
+
+if __name__ == '__main__':
+    cherry_pick_cli()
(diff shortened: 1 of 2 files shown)


Mariatta:
Thanks @lulouie :)

Issue #89, opened 2017-05-13
Change cherry-picker's README to treat it as a package


brettcannon:
If you look at the README for cherry-picker, the instructions tell you to treat it as a module, i.e. `python3 -m cherry-picker` while in the `cherry-picker` directory. But everything else about the code wants it  to be treated as a package, e.g. `__main__.py` uses relative imports so you can't do `python3 .` to execute  cherry-picker.

If cherry-picker is going to be a package then I think the instructions should probably be updated to uniformly treat it as such.

Issue #50, opened 2017-03-17
Add a setup command to cherry_picker


brettcannon:
The idea is to simplify the setup by having `cherry_picker` do the checking out of one's fork and setting the origin. It's minor, but it makes getting started easier and it means that if any steps change in the future then people can be unaware as they just need to remember the setup command instead of assuming they know what to do and missing the new step(s).

Mariatta:
I'll look into it. In the meantime, there is a hidden feature that lets you run cherry_picker.py from an existing cpython repo. Maybe it would have helped..
It was introduced in https://github.com/Mariatta/chic_a_cherry_picker/pull/12 by @ncoghlan just needs to be documented  :stuck_out_tongue: 
I prefer keeping a separate checkout just for cherry-picking just so it won't mess up people's real work.

ncoghlan:
https://github.com/python/core-workflow/pull/53 adds some docs around how the script relates to your git remote config (since the way I typically set things up is the exact opposite of the approach we recommend in the developer guide)

However, I do think a setup command would be useful, as that will move us another step closer to being able to automate the whole flow based on a webhook.

PR #99, opened 2017-05-24
Add setup.py

tiran:
``venv/bin/pip install .`` installs all dependencies and
venv/bin/cherry_picker script.

Signed-off-by: Christian Heimes <christian@python.org>

Closes https://github.com/python/core-workflow/issues/50
Closes https://github.com/python/core-workflow/issues/64
Closes https://github.com/python/core-workflow/issues/89

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,6 +17,8 @@ setup(
     url='https://github.com/python/core-workflow',
     author='Mariatta Wijaya',
     author_email='mariatta.wijaya@gmail.com',
+    maintainer='Python Core Developers',
+    maintainer_email='core-workflow@mail.python.org',
     packages=[
         'cherry_picker',
     ],
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,33 @@
+from setuptools import setup
+
+
+with open('cherry_picker/readme.rst') as f:
+    long_description = f.read()
+
+
+with open('cherry_picker/requirements.txt') as f:
+    requirements = [l.strip() for l in f.read().split('\n')]
+
+
+setup(
+    name='cherry_picker',
+    description='backport cpython changes from master to maintenance branches',
+    long_description=long_description,
+    license='Apache',
+    url='https://github.com/python/core-workflow',
+    author='Mariatta Wijaya',
+    author_email='mariatta.wijaya@gmail.com',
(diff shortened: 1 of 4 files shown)


Mariatta:
Thanks @tiran I believe this will close https://github.com/python/core-workflow/issues/64 So I edited the description.

Will actually review later this evening :)

tiran:
@Mariatta I have amended the commit and updated the readme as request. I also replaced ``(venv) $ python -m cherry_picker`` with just ``(venv) $ cherry_picker``.

Mariatta:
Thanks @tiran :)
I tested this PR, and made some adjustment to the documentation.

@brettcannon or @ncoghlan: Could one of you review this too please?
I'm trying to simplify the setup process, and making the assumption that most people will prefer to cherry_pick from their existing cloned CPython directory instead of cloning it again. So I updated that part of the readme. Thanks.

setup.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+setup(
+    name='cherry_picker',
+    description='backport cpython changes from master to maintenance branches',
+    long_description=long_description,
+    license='Apache',
+    url='https://github.com/python/core-workflow',
+    author='Mariatta Wijaya',
+    author_email='mariatta.wijaya@gmail.com',
```
ncoghlan: You may want to put `core-workflow@gmail.com` as your address here, so folks know to contact the list with any questions, rather than contacting you directly.

tiran: I added @Mariatta 's mail address. We may want to keep it and add maintainer / maintainer_email for contact. After all Marietta wrote the script and deserves credits. :)

Mariatta: Thanks 😸  Well I've prepared a canned response already "Please file an issue at python/core-workflow" 

I'll add the following:
```
"maintainer": "Python Core Developers",
"maintainer_email": "core-workflow@mail.python.org",
```

Question: is `version` metadata not needed? The [doc](https://docs.python.org/3.6/distutils/setupscript.html#meta-data) seems to indicate that it's required.

Mariatta:
Thanks @tiran
You closed 5 open issues with one PR! :tada: :1st_place_medal:

PR #104, opened 2017-06-02
Fix a couple of issues when using the cherry-picker:

Carreau:
- my remotes are over ssh://  which was not handled.
- my remote don't have the .git extension .. for whatever reason.
- --push REMOTE was conflicting with --push/--no-push so I renamed the
  former one to --remote and updated the readme accordingly.

I believe @Mariatta wrote most of that, so your review would be welcome.


Mariatta:
Sorry for all the bugs 😞  and thanks for fixing them 🎉

Carreau:
No need to apologize, I have a pretty weird workflow ! Thanks for writing
things in a first place!

Issue #103, opened 2017-06-01
Error following new cherry_picker setup instructions


csabella:
I started over by deleting the core-workflow directory and then following the setup section.  When I go into the venv and run the first command, I get the following log.  Note, I was able to continue with the steps after this and do a backport.
```
(venv) ~/core-workflow$ python -m pip install --upgrade .
Processing /home/cheryl/core-workflow
Collecting appdirs==1.4.2 (from cherry-picker===0.0.0)
  Using cached appdirs-1.4.2-py2.py3-none-any.whl
Collecting click==6.7 (from cherry-picker===0.0.0)
  Using cached click-6.7-py2.py3-none-any.whl
Collecting packaging==16.8 (from cherry-picker===0.0.0)
  Using cached packaging-16.8-py2.py3-none-any.whl
Collecting py==1.4.33 (from cherry-picker===0.0.0)
  Using cached py-1.4.33-py2.py3-none-any.whl
Collecting pyparsing==2.1.10 (from cherry-picker===0.0.0)
  Using cached pyparsing-2.1.10-py2.py3-none-any.whl
Collecting pytest==3.0.7 (from cherry-picker===0.0.0)
  Using cached pytest-3.0.7-py2.py3-none-any.whl
Collecting pytest-cov==2.4.0 (from cherry-picker===0.0.0)
  Using cached pytest_cov-2.4.0-py2.py3-none-any.whl
Collecting pytest-mock==1.6.0 (from cherry-picker===0.0.0)
  Using cached pytest_mock-1.6.0-py2.py3-none-any.whl
Collecting requests==2.13.0 (from cherry-picker===0.0.0)
  Using cached requests-2.13.0-py2.py3-none-any.whl
Collecting six==1.10.0 (from cherry-picker===0.0.0)
  Using cached six-1.10.0-py2.py3-none-any.whl
Collecting coverage==4.3.4 (from cherry-picker===0.0.0)
  Using cached coverage-4.3.4-cp36-cp36m-manylinux1_x86_64.whl
Collecting setuptools (from pytest==3.0.7->cherry-picker===0.0.0)
  Downloading setuptools-36.0.0-py2.py3-none-any.whl (476kB)
    100% |████████████████████████████████| 481kB 1.3MB/s 
Building wheels for collected packages: cherry-picker
  Running setup.py bdist_wheel for cherry-picker ... error
[...]

PR #105, opened 2017-06-02
Add instructions to README for cherry_picker to install wheel

csabella:
By default, my venv did not have wheel installed, so I needed to install it before running the pip install for cherry_picker.

Closes https://github.com/python/core-workflow/issues/103


Mariatta:
Thanks @csabella 🎉

PR #106, opened 2017-06-02
Rename PR remote option to --pr-remote

ncoghlan:
The new --push/--no-push option conflicts with the old name
of the option for setting the PR remote.

This updates the latter option to the more explicit
`--pr-remote`.


ncoghlan:
Tested by generating https://github.com/python/cpython/pull/1921 and it seems to mostly work.

However, it would be good to remember the PR remote name in the `--continue` state file: https://github.com/python/core-workflow/issues/107 (I haven't added that yet myself, as I'll need to look into how the command continuation works)

Mariatta:
Thanks :)

Issue #108, opened 2017-06-03
Loss of title and description when doing `cherry_picker --continue`


brettcannon:
Maybe I'm doing something wrong, but when I have  used `cherry_picker --continue` I inevitably lose the title and description I would have ended up with had there not been any merge conflict. Not  sure if there's any way to prevent that or to output what the message would have been so I can copy-and-paste it into the eventual PR?

serhiy-storchaka:
This happened to me when I did `git commit` manually after resolving conflicts.

Mariatta:
@brettcannon Did you use `--continue` after a merge conflict, or after using `--no-push`?
And did you have additional commit?

In @serhiy-storchaka's situation, when there are more than one commits in a branch, GitHub will prepopulate the PR title with the branch name. Is this what you meant by losing the title and description?

One way to solve this is to have cherry_picker create the PR using GitHub API / gidgethub, where we can specify the PR title and description ourselves instead of relying on GitHub web UI.

Printing out the original commit message for copy-pasting is the easy way out for now 😛

Mariatta:
In PR #109 the commit message is printed out so you can copy-paste.

brettcannon:
@Mariatta it was with `--continue` after dealing with a merge conflict.

ncoghlan:
This may also align with #107, which we're pretty sure will need a "persistent scratch-pad" where cherry-picker can leave notes for itself in order to remember the original command line options for use with `--continue`.

Probably the simplest option for that would be a JSON file with a config dictionary in it, but if we made it officially an opaque format (e.g. `.cherry-pick.state`), then it could potentially be switched to a full pickle with `shelve` at a later date (and either way it would be easy for other commands to detect that a previous cherry-pick was still in progress due to the state file being present).

webknjaz:
Now with the work done in #295/#277 it should possible to implements storing more state-related data in `git config --local` (i.e. `.git/config`) throughout the whole backport cycle.

PR #109, opened 2017-06-04
Print out commit message after continuing

Mariatta:
Temp fix for #108

### DIFF
--- a/cherry_picker/cherry_picker.py
+++ b/cherry_picker/cherry_picker.py
@@ -245,6 +245,9 @@ To abort the cherry-pick and cleanup:
 
             self.cleanup_branch(cherry_pick_branch)
 
+            click.echo("\nBackport PR:\n")
+            click.echo(updated_commit_message)
+
         else:
             click.echo(u"Currently in `master` branch.  Will not continue. \U0001F61B")
 


cherry_picker/cherry_picker.py
```diff
@@ -245,6 +245,9 @@ def continue_cherry_pick(self):
 
             self.cleanup_branch(cherry_pick_branch)
 
+            click.echo("\nBackport PR:\n")
+            click.echo(updated_commit_message)
```
brettcannon: So is this printing out the title, message body, or both? The previous line is ambiguous without context so I don't know what to expect to see in the eventual output.

Mariatta: This prints out the title, prefixed with [X.Y], message body, and `cherry picked from .. ` message

Example output when backporting PR python/cpython#879:
```
[3.6] bpo-29931 fix __lt__ check in ipaddress.ip_interface for both v4 and v6. (GH-879)                     
                                                                                                            
the original logic was just comparing the network address                                                   
but this is wrong because if the network address is equal then                                              
we need to compare the ip address for breaking the tie                                                      
                                                                                                            
add more ip_interface comparison tests.                                                                     
(cherry picked from commit 7bd8d3e794782582a4ad1c9749424fff86802c3e)  
```

I recorded an asciicast: https://asciinema.org/a/123496

brettcannon: LGTM!

PR #116, opened 2017-06-10
Simplify username parsing

zware:
I've been carrying around a patch that removed the `.git` from the end of `username_end =` line, but it broke with a recent pull, motivating me to fix it properly and submit it.

@Carreau, could you confirm that the `ssh://` URLs I included in the test match what yours look like that you mentioned in #104?


Mariatta:
This LGTM. I'll let you merge it.

zware:
Thanks, @Mariatta!

PR #117, opened 2017-06-10
Make tests verbose on Travis

zware:
`-v` makes it easier to see that any added tests are indeed running as expected.


Mariatta:
Thanks @zware 🎉

Issue #114, opened 2017-06-09
publish cherry_picker on PyPI ?


matrixise:
we have a `setup.py` file, maybe we could use a `core dev` account on PyPI and publish it.

brettcannon:
@dstufft @ncoghlan there isn't some python-dev account that already exists is there?

dstufft:
Not that I'm aware of, but I've never been involved with python-dev publishing anything to PyPI.

Mariatta:
Maybe I'll start by uploading to https://testpypi.python.org/ first 😃  and see how that goes

matrixise:
maybe we could start to publish to PyPI ;-) we are contributors/core dev, maybe we could publish officially our tools. I think it's a shame to clone a repository if we want to use a tool.

just `pip install core-workflow` and we could use the tools.

brettcannon:
So a question for @dstufft for the pip developers account on PyPI: do you have issues with people trying to reset the password since the email address seems to be a public group and thus I would assume anyone could take over the account?

brettcannon:
@matrixise we probably won't do such a broad project as some tools are specific to core devs and others apply to anyone submitting a PR, core dev or not

dstufft:
@brettcannon It probably makes the most sense to use personal accounts unless you're planning on automating the uploads, then use a ``python-dev`` account. In either case, I wouldn't put the password to a public list.

Long term we should have real organization support for PyPI and it could just be owned by an org instead.

Mariatta:
I uploaded cherry_picker to testpypi https://testpypi.python.org/pypi/cherry_picker/ 

:thinking: Do I need to reformat the readme differently?

dstufft:
@Mariatta You must have a rendering error, try:

```console
$ pip install readme_renderer
$ python setup.py check -rs
``

That should give you more information.

Mariatta:
Thanks! 😃 It's rendering better now.

gpshead:
pypi org support would be awesome... but until that happens so long as we have 3+ personal accounts of devs listed as owners that should be sufficient.

brettcannon:
I can be listed as one of the collaborators, @Mariatta , if you want.

Mariatta:
> I can be listed as one of the collaborators, @Mariatta , if you want.

Ok 😄  I just don't know how to do it 😐  I've never uploaded anything to PyPI.

willingc:
[...]

PR #119, opened 2017-06-10
Updates to cherry_picker readme, for PyPI

Mariatta:
Add version info in setup.py
Cleanup readme

Uploaded to testpypi: https://testpypi.python.org/pypi/cherry_picker/

#114


Mariatta:
Setup info can be updated with `pip install cherry_picker` but maybe that's a separate PR

PR #142, opened 2017-06-24
Drop requirements.txt

brettcannon:
Covered by flit.ini.


Mariatta:
🤔 Hmm, it's not needed for packaging purpose, but I still use the requirements file to install the dependencies in my local dev environment. Is there a different way to do it now?
Travis needs it too ...

brettcannon:
You can use [`flit install`](https://flit.readthedocs.io) to install dependencies (use the `--symlink` flag if you want to use symlinks for the install for cherry_picker which will let you edit the files so changes take effect for the installed files).

Mariatta:
Thanks!

PR #146, opened 2017-06-26
Replace f-string with "...".format(...)

haypo:
Change adding Python 3.5 support.

### DIFF
--- a/blurb/blurb.py
+++ b/blurb/blurb.py
@@ -163,10 +163,10 @@ def sortable_datetime():
 
 
 def prompt(prompt):
-    return input(f"[{prompt}> ")
+    return input("[{}> ".format(prompt))
 
 def require_ok(prompt):
-    prompt = f"[{prompt}> "
+    prompt = "[{}> ".format(prompt)
     while True:
         s = input(prompt).strip()
         if s == 'ok':
@@ -393,7 +393,7 @@ class Blurbs(list):
         line_number = None
 
         def throw(s):
-            raise BlurbError(f"Error in {filename}:{line_number}:\n{s}")
+            raise BlurbError("Error in {}:{}:\n{}".format(filename, line_number, s))
 
         def finish_entry():
             nonlocal body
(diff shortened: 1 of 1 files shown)


haypo:
Alternate syntax of PR #143, again, to add Python 3.5 compatibility.

blurb/blurb.py
```diff
@@ -495,10 +495,10 @@ def _parse_next_filename(filename):
         components = filename.split("/")
         section, filename = components[-2:]
         section = unsanitize_section(section)
-        assert section in sections, f"Unknown section {section}"
+        assert section in sections, "Unknown section {}".format(section)
 
         fields = [x.strip() for x in filename.split(".")]
-        assert len(fields) >= 4, f"Can't parse 'next' filename! filename {filename!r} fields {fields}"
+        assert len(fields) >= 4, "Can't parse 'next' filename! filename {!r} fields {}".format(filename, fields)
```
1st1: This line is surely longer than 79 characters.

haypo: The PEP 8 allows up to 120 characters. The file already contains lines longer than 120 characters (before my change) ;-)

serhiy-storchaka:
Thank you for doing this @haypo!

serhiy-storchaka:
Brett, could you please update the version on PyPI? Current version is installable on 3.5, but doesn't work.

brettcannon:
@serhiy-storchaka sure, but I'm traveling ATM so I won't be able to get to it until I return home (no later than sometime next week).

Mariatta:
Prior to releasing to PyPI, I think [this line](https://github.com/python/core-workflow/blob/master/blurb/flit.ini#L12) in flit.ini should be updated so it can work with Python < 3.6.

The readme still states "blurb's only dependency is Python 3.6+.", guess that should be updated too.

brettcannon:
I updated the version number, tweaked the trove classifiers, fixed the README, pushed [1.0.1](https://pypi.org/project/blurb/1.0.1/) to PyPI, and then bumped the version number for future development.

serhiy-storchaka:
Unfortunately
```
python3 -m pip install --user blurb
```
still installs version 1.0. And explicit specifying the version 1.0.1 doesn't work.
```
$ python3 -m pip install --user blurb==1.0.1
Collecting blurb==1.0.1
  Could not find a version that satisfies the requirement blurb==1.0.1 (from versions: 1.0)
No matching distribution found for blurb==1.0.1
```

PR #150, opened 2017-06-28
Set cherry_picker to v 0.1.0.dev1

Mariatta:
`click` doesn't recognize the version set in` __init__.py` 😞  so it has to be hardcoded.

### DIFF
--- a/cherry_picker/cherry_picker/__main__.py
+++ b/cherry_picker/cherry_picker/__main__.py
@@ -1,4 +1,4 @@
-import cherry_pick_cli
+from .cherry_picker import cherry_pick_cli
 
 if __name__ == '__main__':
     cherry_pick_cli()
--- a/cherry_picker/cherry_picker/__main__.py
+++ b/cherry_picker/cherry_picker/__main__.py
@@ -1,4 +1,4 @@
-from .cherry_picker import cherry_pick_cli
+import cherry_pick_cli
 
 if __name__ == '__main__':
     cherry_pick_cli()
--- a/cherry_picker/cherry_picker/cherry_picker.py
+++ b/cherry_picker/cherry_picker/cherry_picker.py
@@ -7,6 +7,9 @@ import subprocess
 import webbrowser
 import sys
 
+from . import __version__
+
+
 class CherryPicker:
 
     def __init__(self, pr_remote, commit_sha1, branches,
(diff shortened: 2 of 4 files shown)


cherry_picker/cherry_picker/cherry_picker.py
```diff
@@ -251,7 +251,7 @@ def continue_cherry_pick(self):
 CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help'])
 
 @click.command(context_settings=CONTEXT_SETTINGS)
-@click.version_option(version="0.0.2.post2")
+@click.version_option(version="0.1.0.dev1")
```
zware: Why not `from . import __version__`?  Or the other way around (define it here, import it in `__init__.py`), if flit supports that?

I'm not really clear on why exactly `__init__.py` is necessary at all :)

willingc: I was thinking the same as @zware. Is it a click or a flit limitation?

Mariatta: Hmm.. it's an old artifact :stuck_out_tongue_closed_eyes: 
Thanks :) I'll remove `__init__.py` and move `__version__` to `cherry_picker.py`

Mariatta: 🤔 When I removed __init__.py, getting this error message during `flit install`:
```
FileNotFoundError: [Errno 2] No such file or directory: 'cherry_picker/__init__.py'
```
So I kept the file..

cherry_picker/cherry_picker/__main__.py
```diff
@@ -1,4 +1,4 @@
-from .cherry_picker import cherry_pick_cli
+import cherry_pick_cli
```
zware: This doesn't seem right; `cherry_pick_cli` isn't a module, is it?

cherry_picker/flit.ini
```diff
@@ -2,6 +2,8 @@
 module = cherry_picker
 author = Mariatta Wijaya
 author-email = mariatta.wijaya@gmail.com
+maintainer = Python Core Developers
+maintainer-email = core-workflow@mail.python.org
```
zware: I think the `mail.` is superfluous.

Mariatta: Thanks :) Removed `mail.`. It was copied from [blurb](https://github.com/python/core-workflow/blob/master/blurb/flit.ini#L6) but I'll leave that alone for now :)

Mariatta:
Thanks :)

willingc:
Thanks 💐

Mariatta:
Question... how do I make it the default version in https://pypi.org/project/cherry-picker/  ?
It's still showing v 0.0.2.post1 by default even though 0.1.0 is available..

PR #155, opened 2017-07-07
Add vi to fallbacks list

### LINKED ISSUE #154: blurb refuse to work without setting EDITOR env
blurb is too picky:
```
$ python3 /home/nsoffer/src/core-workflow/blurb/blurb.py 
Error: Could not find an editor! Set the EDITOR environment variable.
```

Git is happy on this system using vim, why blurb has to be so picky?

nirs:
If EDITOR or GIT_EDITOR is not set, we use a fallback editor, but on
non-windows only "nano" was listed, and it is not installed by default
on recent Linux distributions, while vi is practically everywhere.

Fixes #154

### DIFF
--- a/blurb/blurb.py
+++ b/blurb/blurb.py
@@ -752,7 +752,7 @@ def find_editor():
     if sys.platform == 'win32':
         fallbacks = ['notepad.exe']
     else:
-        fallbacks = ['/etc/alternatives/editor', 'nano']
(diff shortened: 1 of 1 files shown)


larryhastings:
Nope.  vi is not an acceptable editor for new users, and it's only new users who won't have a default EDITOR set anyway.

PR #157, opened 2017-07-08
Abort "blurb add" cleanly

nirs:
If the user aborted the operation by hitting Control+C, abort the
operation without a traceback.

Example run:

    $ blurb
    (Terminate editor without any change)

    Error: Error in /tmp/tmpt0etkaf2.rst:26:
    Blurb 'body' text must not be empty!

    [Hit return to retry (or Ctrl-C to abort)> ^C
    $


larryhastings:
Thanks!  Merged.

PR #158, opened 2017-07-08
update blurb's flit.ini, it requires Python >= 3.5

Mariatta:


brettcannon:
I actually did the change without seeing this PR, but merging a redundant PR doesn't hurt either. 😉

PR #161, opened 2017-07-21
blurb: small usability improvements

JelleZijlstra:
These would have made my first experience using blurb just now smoother.
- Make `blurb --help` work in addition to `blurb help`. This is my default command for figuring out how to use a command; for example, `mypy --help`, `pip --help`, and `flit --help` all give me useful info.
- Suggest looking at `blurb help` if the user gave an invalid command.

### DIFF
--- a/blurb/blurb.py
+++ b/blurb/blurb.py
@@ -677,7 +677,7 @@ If subcommand is not specified, prints one-line summaries for every command.
         summaries = []
         longest_name_len = -1
         for name, fn in subcommands.items():
-            if name == '--help':
+            if name.startswith('-'):
                 continue
             longest_name_len = max(longest_name_len, len(name))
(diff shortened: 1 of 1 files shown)


blurb/blurb.py
```diff
@@ -677,6 +677,8 @@ def help(subcommand=None):
         summaries = []
         longest_name_len = -1
         for name, fn in subcommands.items():
+            if name == '--help':
```
larryhastings: Can you change this to `if name.startswith("-"):` ?  That would let us support `--version`` too.

JelleZijlstra: Done. (There isn't currently a "version" subcommand though.)

larryhastings:
Sorry this languished for so long!

Issue #100, opened 2017-05-24
Automate backport PR creation


Mariatta:
Maybe we can use this to automate the PR creation in `cherry_picker.py`: 
https://github.com/jd/git-pull-request 

Blog post: 
https://julien.danjou.info/blog/2017/git-pull-request-command-line-tool 

Update:
Use gidgethub instead. For authentication, we'll use a personal access token, specified in the environment variable.

Mariatta:
Hmm but this requires people to write their GitHub credentials to write your GitHub credentials to `~/.netrc` file, as mentioned in their readme:

```
machine github.com login jd password f00b4r
```


Not sure if users will be comfortable with this idea?

matrixise:
Normally the .netrc file is readonly for the owner but maybe there is the capability to use a token, we can ask to Julien, it's a core dev of OpenStack at Redhat and I know him and also @haypo

Mariatta:
Thanks @matrixise :)

I guess in the end it comes to the question of how we can store the password / token securely. 
We can also use Brett's gidgethub to create the PR, the question is still: how to we store the token/credential securely? 

@brettcannon how do you store the credentials when gidgethub is used for ni or bedevere?

jd:
There are plenty of other options, such as using keyrings (OS X or GNOME have that as builtin), token etc.

I took the `.netrc` approach as it's easy and standard, but the code should make it pretty easy to add any new better/easier mechanism!

Issue #8, opened 2016-12-10
Bot to automatically generate cherry-picking PRs


brettcannon:
As [pulled from PEP 512](https://www.python.org/dev/peps/pep-0512/#bot-to-generate-cherry-pick-pull-requests):

Since the decision has been made to work with cherry-picks instead of forward merging of branches, it would be convenient to have a bot that would generate pull requests based on cherry-picking for any pull requests that affect multiple branches. The most likely design is a bot that monitors merged pull requests with key labels applied that delineate what branches the pull request should be cherry-picked into. The bot would then generate cherry-pick pull requests for each label and remove the labels as the pull requests are created (this allows for easy detection when automatic cherry-picking failed).

soltysh:
I'm interested in helping with this one 😄

brettcannon:
Yeah, I think this will be one of those high-impact improvements that will definitely make PR acceptance more pleasurable compared to the hg workflow. I think this should be number 2 on the TODO list once #6 is dealt with (as I see the NEWS file being the stickiest point in accepting PRs and it also makes this issue easier by lowering potential merge conflicts).

berkerpeksag:
Note that I already have a working bot (tested it on https://github.com/fayton/cpython), but I didn't had time to clean it up and make it open source in the past months.

My plan is to work on bug fixes for 3.5.3, 3.6.1 and python.org over the holiday. I can help you after mid-January if you had time to create a prototype during the holiday (or you can help me to clean up my bot if none of us can work on a prototype before mid-January -- I codenamed it "Duke Nukem Forever" :))

soltysh:
👍 @berkerpeksag we can sync and I'd love to help with that.

berkerpeksag:
Just an FYI, I've finally cleaned-up my bot and set up on my test repository. There is a missing feature though: it doesn't create a new pull request; it automatically pushes the cherry-picked commit into maintenance branches.

See https://github.com/fayton/cpython/pull/2/#issuecomment-278313431 for an example. Note that it also automatically fills the committer field by inspecting the "core developers" team created on the organization.

brettcannon:
What about if the cherry-pick fails either due to testing or the code not patching cleanly? And how do you keep track of what has (not) successfully been cherry-picked (I'm assuming some label on the original PR)?

Carreau:
> What about if the cherry-pick fails either due to testing or the code not patching cleanly?

[...]

PR #175, opened 2017-08-27
Automatically create backport PR if oauth token is set

Mariatta:
- Add `gidgethub` and `requests` as dependencies.

Closes https://github.com/python/core-workflow/issues/100


Mariatta:
https://github.com/python/cpython/pull/3224 was created via this method, using my own personal access token.

terryjreedy:
I have read the readme, but it is not clear to me what 'create automatically' means -- what will happen, when and where -- and what a token on my machine has to do with it.

Mariatta:
Thanks @terryjreedy. The idea is instead of opening a web browser where we have to click the 'Create pull request button', cherry picker will create the PR. So no web browser will be opened at all.
cherry_picker will print out the URL of the newly created pull request.

The token is needed by GitHub API as a way to authenticate your GitHub account.

terryjreedy:
Thanks for explaining.  Since I like to have the PR open in my browser, and clicking the button is an easy way to get that, I will probably pass.

Mariatta:
No prob :) It's backward compatible, so the previous way still works as is.

Potentially we can use this for the cherry-picking bot, where the backport will happen automatically after we merge the PR on `master`. (Issue https://github.com/python/core-workflow/issues/8)

cherry_picker/readme.rst
```diff
@@ -53,6 +57,19 @@ repository are pushed to ``origin``. If this is incorrect, then the correct
 remote will need be specified using the ``--pr-remote`` option (e.g.
 ``--pr-remote pr`` to use a remote named ``pr``).
 
+Personal Access Token (optional)
+--------------------------------
+
+``cherry_picker`` can automatically create the PR, if the ``GH_AUTH`` token is set
```
brettcannon: Is this considered safe and best practice for storing an auth token like this? I honestly don't know, which is why I'm asking. 😄

Mariatta: I don't really know 😅 
This is really so I can use it for the cherry picker bot later.
Maybe I should keep this as hidden and secret functionality, and not mention it at all in the readme?

brettcannon: Yeah, either don't document it or be clear that one should be careful about how that environment variable is stored.

Mariatta: Ok I'm going with "not documenting it" since it's easier to do than coming up with warning messages 😅

Mariatta:
I didn't expect the Spanish Inquisition!

cherry_picker/cherry_picker/cherry_picker.py
```diff
@@ -6,9 +6,13 @@
 import subprocess
 import webbrowser
 import sys
+import requests
+
+from gidgethub.sansio import create_headers
```
brettcannon: Not important, but you shifted from importing modules to importing objects with this import line.

cherry_picker/cherry_picker/cherry_picker.py
```diff
@@ -150,7 +155,27 @@ def push_to_remote(self, base_branch, head_branch):
         except subprocess.CalledProcessError:
             click.echo(f"Failed to push to {self.pr_remote} \u2639")
         else:
-            self.open_pr(self.get_pr_url(base_branch, head_branch))
+            if os.getenv("GH_AUTH"):
+                self.create_gh_pr(base_branch, head_branch, commit_message)
+            else:
+                self.open_pr(self.get_pr_url(base_branch, head_branch))
+
+    def create_gh_pr(self, base_branch, head_branch, commit_message):
```
brettcannon: If you're already checking for the envvar above, would it make sense to make it a required argument and just pass it in explicitly?

Mariatta: Ok, I made the token a required argument.

cherry_picker/cherry_picker/cherry_picker.py
```diff
@@ -8,7 +8,7 @@
 import sys
 import requests
 
-from gidgethub.sansio import create_headers
+import gidgethub
```
brettcannon: This should either be `import gidgethub.sansio` or `from gidgethub import sansio`; the fact that the package has an import is undocumented.

cherry_picker/cherry_picker/cherry_picker.py
```diff
@@ -155,13 +155,19 @@ def push_to_remote(self, base_branch, head_branch, commit_message=""):
...
+            gh_auth = os.getenv("GH_AUTH")
+            if gh_auth:
+                self.create_gh_pr(base_branch, head_branch, commit_message,
+                                  gh_auth)
             else:
                 self.open_pr(self.get_pr_url(base_branch, head_branch))
 
-    def create_gh_pr(self, base_branch, head_branch, commit_message):
-        request_headers = create_headers(self.username, oauth_token=os.getenv("GH_AUTH"))
+    def create_gh_pr(self, base_branch, head_branch, commit_message, gh_auth):
```
brettcannon: Is there any worry the order of these arguments will be non-obvious? I.e. should any of them be keyword-only arguments?

Mariatta: Hmm, ok I made commit_message and gh_auth as keyword only arguments.

cherry_picker/cherry_picker/cherry_picker.py
Mariatta: Changed to`from gidgethub import sansio`

PR #178, opened 2017-09-04
Allows blurb to correctly split paths on Windows

zooba:


blurb/blurb.py
```diff
@@ -519,7 +519,7 @@ def _parse_next_filename(filename):
 Parses a "next" filename into its equivalent blurb metadata.
 Returns a dict.
         """
-        components = filename.split("/")
+        components = filename.split(os.sep)
```
zware: Should we also `os.path.normpath`?

zooba: They're all obtained via glob, so it isn't going to be necessary.

zooba:
@larryhastings This is the change - it's been merged, so just need the `setup.py sdist upload` step

larryhastings:
Well, one more step needed: I had to increment the version number.  blurb 1.0.5 is now pushed to the cheeseshop.

PR #184, opened 2017-09-20
blurb: get BPO# from branch, add commit template

tiran:
blurb now gets the branch name from git and auto-fills the bpo number if
the branch name looks like "bpo-1234-whatever". "bpo1234" or
"bpo_1234_whatever" are also supported.

Further more blurb now writes .git/blurb commit template. The first line
of the template is "bpo-{bpo}: suffix of branch name". The body of the
template is the blurb text.

```
$ git config commit.template .git/blurb

$ git checkout -b bpo-1234-confuse-a-cat
$ blurb
git ci
---
bpo-1234: confuse a cat

blurb text
---
```

Signed-off-by: Christian Heimes <christian@python.org>

### DIFF
--- a/blurb/blurb.py
+++ b/blurb/blurb.py
@@ -75,7 +75,7 @@ template = """
 #
 # Please enter the relevant bugs.python.org issue number here:
 #
-.. bpo:
+.. bpo: {bpo}
 
 #
 # Uncomment one of these "section:" lines to specify which section
@@ -838,24 +838,14 @@ Add a blurb (a Misc/NEWS entry) to the current CPython repo.
     os.close(handle)
     atexit.register(lambda : os.unlink(tmp_path))
 
-    def init_tmp_with_template():
-        with open(tmp_path, "wt", encoding="utf-8") as f:
-            # hack:
-            # my editor likes to strip trailing whitespace from lines.
-            # normally this is a good idea.  but in the case of the template
-            # it's unhelpful.
-            # so, manually ensure there's a space at the end of the bpo line.
-            text = template
+    # see of the branch name indicated BPO number
+    # makes the 'editor strips trailing whitespace from template lines'
+    # hack obsolete, too.
+    branch_bpo, branch_suffix = get_bpo_git_branch()
 
-            bpo_line = ".. bpo:"
-            without_space = "\n" + bpo_line + "\n"
-            with_space = "\n" + bpo_line + " \n"
-            if without_space not in text:
-                sys.exit("Can't find BPO line to ensure there's a space on the end!")
-            text = text.replace(without_space, with_space)
-            f.write(text)
-
-    init_tmp_with_template()
+    with open(tmp_path, "wt", encoding="utf-8") as f:
+        text = template.format(bpo=branch_bpo)
+        f.write(text)
 
     while True:
         subprocess.run([editor, tmp_path])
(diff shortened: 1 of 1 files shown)


Mariatta:
I like this feature, but for it to be effective, people need to actually create their branch as "bpo-NNNN". 
If Larry accepts this PR, then it should be followed with an update to the devguide to recommend a branch name of "bpo-NNNN".

blurb/blurb.py
```diff
@@ -838,24 +838,14 @@ def add():
...
-            text = template
-
-            bpo_line = ".. bpo:"
-            without_space = "\n" + bpo_line + "\n"
-            with_space = "\n" + bpo_line + " \n"
-            if without_space not in text:
-                sys.exit("Can't find BPO line to ensure there's a space on the end!")
-            text = text.replace(without_space, with_space)
-            f.write(text)
+    # see of the branch name indicated BPO number
```
larryhastings: I realize English is not your first language; may I propose "see if the branch name indicates the BPO number"?

blurb/blurb.py
```diff
@@ -838,24 +838,14 @@ def add():
...
+    # makes the 'editor strips trailing whitespace from template lines'
```
larryhastings: Given that your patch obviates (and subsequently removes) the hack in question, this two-line comment is meaningless.

blurb/blurb.py
```diff
@@ -888,6 +878,18 @@ def init_tmp_with_template():
     path = blurb.save_next()
     git_add_files.append(path)
     flush_git_add_files()
+
+    # For convenience, write a preformatted commit template. Enable template
```
larryhastings: Again, a minor suggestion regarding your English here.  The first sentence is fine, for the second may I suggest "You can tell git to use this template with "git  config commit.template .git/blurb"."

It's not a big deal, but I don't know why you use `` to quote the command here.  Simple double-quotes are fine, and if you want something clearer I would indent and put a % to indicate a command prompt.

Also, based on the code, it looks like

1. this only works if the ~/.git/blurb directory exists, and
2. blurb doesn't create the directory for you automatically.

Does git create the directory for you?  If not, we should probably suggest that _too_ in the comment, or else people will follow the instructions only to discover that they don't work.

blurb/blurb.py
```diff
@@ -888,6 +878,18 @@ def init_tmp_with_template():
...
+    # with ``git config commit.template .git/blurb``.
+    msgfile = os.path.join(root, '.git', 'blurb')
+    if os.path.isdir(os.path.dirname(msgfile)):
+        metadata, text = blurb[0]
+        title = branch_suffix.replace('-', ' ').replace('_', ' ')
```
larryhastings: I don't like this.  You are replacing _all_ dashes and _all_ underscores with spaces here.  This means that a branch called `"bpo-1234-Fix-__dir__-on-dict"` would be unhelpfully converted into the template `"    bpo-1234: Fix   dir   on dict"` with _three_ spaces on either side of `dict`.  (GitHub apparently refuses to display the three spaces, 
and collapses them to one during formatting.)

If we honor the name of the branch, we should honor the name of the branch.  There are legitimate uses for underscores and dashes, and stripping them all out is more surprising than simply leaving them there.  I realize git doesn't permit spaces in branch names, so we can't simply write a wonderful commit message in the branch name and have it emerge from the other end of the process perfectly.  Ultimately, leaving the dashes and underscores there may be a little inelegant but that's better than outright corrupted.

Please remove this line and resubmit the PR.

blurb/blurb.py
```diff
@@ -1067,6 +1069,29 @@ def flush_git_rm_files():
...
+    returns: branch number, branch suffix
+    """
+    try:
+        p = subprocess.run(["git", "rev-parse", "--abbrev-ref", "HEAD"], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        p.check_returncode()
+    except subprocess.CalledProcessError:
+        return '', ''
+    branch = p.stdout.decode('utf-8').strip()
+    mo = re.match("^bpo[-_]?(\d+)[-_]?(.*)", branch)
+    if mo is not None:
```
larryhastings: Eschew The Unnecessary Else!

I propose you rewrite these four lines as:

```
   if mo is None:
      return '', branch
   return mo.group(1), mo.group(2)
```

blurb/blurb.py
terryjreedy: Our current PEP 8 standard for comments is to make them complete sentences, with initial cap and period, as in
`# See if the branch name indicates the BPO number.`
Many old comments do not meet this standard, but new or edited ones should.

matrixise:
@tiran I am interested in this feature, do you need help for this code?

Thank you

matrixise:
@Mariatta I have updated this PR with that one https://github.com/python/core-workflow/pull/308

hugovk:
Closing this in favour of the updated https://github.com/python/core-workflow/pull/308.

Issue #164, opened 2017-07-24
Create a list of tools we use with the new workflow


berkerpeksag:
It would be nice to gather all the tools we use in a list so we can easily report bugs or feature requests without having to search on Google or list archives. Plus, future contributors can easily find them if they are looking for an alternative area to work on.

Suggested tools:

* https://github.com/python/bedevere
* https://github.com/python/the-knights-who-say-ni
* https://github.com/berkerpeksag/cpython-emailer-webhook
* https://github.com/berkerpeksag/cpython-bpo-linkify

Suggested format:

Name | Description | Issue tracker | Owner/Maintainer
---- | ----------- | ------------- | ----------------
[python/bedevere](https://github.com/python/bedevere) | A bot to help identify missing information for CPython pull requests | [GitHub](https://github.com/python/bedevere/issues) | [Brett Cannon](https://github.com/brettcannon)

Mariatta:
Do you think the above list can be added here in the readme of core-workflow, or in the dev guide?

brettcannon:
I think the README works; don't see why someone wanting to contribute to CPython is explicitly going to care about where the tools come from, just how to use them.

berkerpeksag:
Yes I agree, `core-workflow/README.rst` would be a better place to put the list.

pablogsal:
I am so sorry for the spam. Clearly, GitHub doesn't like `git --amend' of commits that refer to issues.

PR #185, opened 2017-10-03
Create a list of tools we use with the new workflow. Solves #164

pablogsal:
This PR creates a (preliminary) list of core workflow tools 🔧  as indicated by #164. 

Please, if some other core-workflow tools are missing, indicate it and I will be more than happy to add them. 😃 

The rendering of the `rst` file can be checked [here](https://github.com/pablogsal/core-workflow/blob/issue-164/README.rst).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -34,3 +34,20 @@ Interactive utility for writing CPython ``Misc/NEWS.d`` entries. See
 the blurb_ directory for more details.
 
 .. _blurb: https://github.com/python/core-workflow/tree/master/blurb
+
+
+Other core workflow tools
+-------------------------
+
++------------------------------------------------------------------------------------------------------+--------------------------------------------------------------------------------------------+-------------------------------------------------------------------------------+-------------------------------------------------------+
+| Name                                                                                                 | Description                                                                                | Issue tracker                                                                 | Owner/Maintainer                                      |
++======================================================================================================+============================================================================================+===============================================================================+=======================================================+
+| `python/bedevere <https://github.com/python/bedevere>`__                                             | A bot to help identify missing information for CPython pull requests                       | `Github <https://github.com/python/bedevere/issues>`__                        | `Brett Cannon <https://github.com/brettcannon>`__     |
++------------------------------------------------------------------------------------------------------+--------------------------------------------------------------------------------------------+-------------------------------------------------------------------------------+-------------------------------------------------------+
+| `python/the-knights-who-say-ni <https://github.com/python/the-knights-who-say-ni>`__                 | CLA enforcement bot for Python organization projects                                       | `Github <https://github.com/python/the-knights-who-say-ni/issues>`__          | `Brett Cannon <https://github.com/brettcannon>`__     |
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -34,3 +34,20 @@ Interactive utility for writing CPython ``Misc/NEWS.d`` entries. See
 the blurb_ directory for more details.
 
 .. _blurb: https://github.com/python/core-workflow/tree/master/blurb
+
+
+Other core workflow tools
+-------------------------
+
++------------------------------------------------------------------------------------------------------+--------------------------------------------------------------------------------------------+-------------------------------------------------------------------------------+-------------------------------------------------------+
```
berkerpeksag: I think we can use the simpler syntax at http://docutils.sourceforge.net/docs/user/rst/quickref.html#tables

berkerpeksag: Also, I might be wrong, but only the first column of a row should match the length of its header. For example, the following table is valid in reST:

```rst
============== =============
Name           Issue tracker
============== =============
Some long name This is a very long URL of the issue tracker
Some long name https://www.python.org/
============== =============
```

That would simplify markup of the table. See the latest example at http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#simple-tables for more details.

README.rst
```diff
@@ -34,3 +34,20 @@ Interactive utility for writing CPython ``Misc/NEWS.d`` entries. See
...
+| Name                                                                                                 | Description                                                                                | Issue tracker                                                                 | Owner/Maintainer                                      |
++======================================================================================================+============================================================================================+===============================================================================+=======================================================+
+| `python/bedevere <https://github.com/python/bedevere>`__                                             | A bot to help identify missing information for CPython pull requests                       | `Github <https://github.com/python/bedevere/issues>`__                        | `Brett Cannon <https://github.com/brettcannon>`__     |
```
berkerpeksag: Nitpick: Github -> GitHub

README.rst
```diff
@@ -34,3 +34,20 @@ Interactive utility for writing CPython ``Misc/NEWS.d`` entries. See
...
++------------------------------------------------------------------------------------------------------+--------------------------------------------------------------------------------------------+-------------------------------------------------------------------------------+-------------------------------------------------------+
+| `python/the-knights-who-say-ni <https://github.com/python/the-knights-who-say-ni>`__                 | CLA enforcement bot for Python organization projects                                       | `Github <https://github.com/python/the-knights-who-say-ni/issues>`__          | `Brett Cannon <https://github.com/brettcannon>`__     |
```
berkerpeksag: You can also do the following to avoid creating duplicate hyperlink markup:

```rst
`Brett Cannon`_

.. You can now just use `Brett Cannon`_ to link to their GitHub profile in multiple places:

`Brett Cannon`_ is a Python core developer.

.. _`Brett Cannon`: https://github.com/brettcannon
```

pablogsal:
All feedback addressed on 59bebca. ^^

berkerpeksag:
Thanks!

PR #190, opened 2017-10-17
Add -f to blurb to include files even if ignored

mariocj89:
Users might have a .gitignore configuration where they ignore folders like "Lib", "Build" "Misc" or other paths that might lead to blurb not being able to add the files using just "git add". This change makes git to "force" the addition of the files passed in, ensuring that even if present in gitignore they will be added to the commit, rather than blurb just failing in the operation.

Sending the patch as it just happened to me 😄

### DIFF
--- a/blurb/blurb.py
+++ b/blurb/blurb.py
@@ -1046,7 +1046,7 @@ Python News
 git_add_files = []
 def flush_git_add_files():
     if git_add_files:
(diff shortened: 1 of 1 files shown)


larryhastings:
I don't understand.  CPython's .gitignore doesn't ignore "Misc", 

Is there a second user-settable ".gitignore" which git will also, simultaneously, obey?

mariocj89:
In git you can have a global .gitignore 

Configurable via `git config --global core.excludesfile ~/.gitignore_global`

larryhastings:
Thanks for the PR!

PR #203, opened 2017-12-07
Test with Python 3.7-dev

Mariatta:


brettcannon:
Actually, is this necessary if we are already testing against nightly?

PR #213, opened 2018-01-31
Improve the way we retrieve the full commit hash from shorthash

Mariatta:


Mariatta:
Thanks for catching the typo! :)

PR #236, opened 2018-04-12
Fix readme: toml requires lower-cased true/false for bools

asvetlov:
Update cherry-picker config example.


asvetlov:
Sorry, my bad.

cherry_picker/readme.rst
```diff
@@ -333,7 +333,7 @@ Changelog
 - Add ``fix_commit_msg`` configuration item. Setting fix_commit_msg to ``True``
```
serhiy-storchaka: ``` ``true`` ```?

asvetlov: Yes, thank you.

PR #248, opened 2018-05-22
blurb: don't break after hyphens and too long words.

### LINKED ISSUE #247: blurb can break a line after a hyphen
Sphinx interprets the linebreak as a whitespace and adds a space after a hyphen in the rendered text.

For examples see https://github.com/python/cpython/pull/7002.

There is an option for `textwrap` that disables breaking lines on hyphens.

serhiy-storchaka:
Fixes #247


serhiy-storchaka:
@ned-deily, please use the patched blurb for the next release. Otherwise it will produce broken news files.

PR #252, opened 2018-05-23
Refactor Travis CI config

webknjaz:
It's better readable this way and it's also much easier to distinguish between failed things when different things are tested in different jobs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,32 +18,21 @@ install:
 script:
 - blurb test
 
+.mixtures:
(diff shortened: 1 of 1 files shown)


webknjaz:
@asvetlov right, and I took this into account. blurb runs 3.5->nightly. cherry_picker runs 3.6->nightly.

webknjaz:
@asvetlov anything else concerning you?

Mariatta:
Looks great! Thanks!

Issue #222, opened 2018-02-13
Automatically deploy cherry_picker (and/or blurb) to PyPI


Mariatta:
I heard that this can be accomplished using TravisCI or CircleCI.

Mariatta:
TravisCI docs: https://docs.travis-ci.com/user/deployment/script/ 
Might not work since cherry_picker now uses pyproject.toml instead of setup.py

CircleCI example https://circleci.com/blog/continuously-deploying-python-packages-to-pypi-with-circleci/

Mariatta:
My own idea is to write a bot that does `flit publish`.

webknjaz:
I can do it. If needed, I'll send a PR to travis-ci/dpl for the existing pypi provider (I wrote Pages provider, so I'm familiar with the tool), but it's likely that it won't be necessary.

Another thing to consider would be the deployment flow trigger. Normally, I would do deploy for every tag pushed and even use that version-tag as a version of current distribution (see setuptools_scm).
But in this repo there's two distributions (blurb and cherry-picker), which have different versions, so we need some hack to distinguish them.
I suggest something like prefixing `cherry-picker-vx.x.x` or `blurb-vy.y.y` must be enough. If you don't wont to rely on git tags in dist you'll still have to edit and commit version in source and then tag it (as opposed to just tagging and pushing a commit).

Regarding bot, I've got some ideas and having ongoing investigation of how GitHub Deployments APIs work along with GitHub Apps (login as a bot) + OAuth there. I don't have any public demo yet, but the plan is to make some UI with a button, which would trigger a deployment event and then the event listener part of app will react on such events and do whatever needed. I'm also considering another scenario where CIs will send such events. If I understand correctly, one may even add a condition into such trigger in API to wait for all CIs to complete, which is going to be very useful for aio-libs (where I originally intended to apply this), because in case of several CIs building OS-specific wheels asynchronously the wheels appear in PYPI during 8-hour interval, which seems to be very confusing to users. So this bot could be a nice "synchronization point" for such processes.

But it all is just a theory for now. I'll start with what I know best: built-in deployment provider.

webknjaz:
@Mariatta Please let me know if you prefer somewhat different flow. And I guess we could move the discussion to #262, shall we?

PR #253, opened 2018-05-23
Support version retrieval from custom branches

### LINKED ISSUE #251: [cherry-picker] Misleading assumption about backport branches naming
While it's true for cpython and aiohttp, not everyone uses just version numbers for their backport branches.

Someone may use v-prefix `v2.6`, we use `stable-2.6`. There might be other cases.
I suggest changing how version is extracted from the branch name, so that it would support more generic flows.

webknjaz:
Fixes #251

### DIFF
--- a/cherry_picker/cherry_picker/cherry_picker.py
+++ b/cherry_picker/cherry_picker/cherry_picker.py
@@ -7,6 +7,7 @@ import os
 import pathlib
 import subprocess
 import webbrowser
+import re
 import sys
 import requests
 import toml
@@ -72,10 +73,16 @@ class CherryPicker:
 
     @property
     def sorted_branches(self):
+        def version_from_branch(branch):
+            try:
+                return tuple(map(int, re.match(r'^.*(?P<version>\d+(\.\d+)+).*$', branch).groupdict()['version'].split('.')))
+            except AttributeError as attr_err:
+                raise ValueError(f'Branch {branch} seems to not have a version in its name.') from attr_err
+
         return sorted(
             self.branches,
             reverse=True,
-            key=lambda v: tuple(map(int, v.split('.'))))
+            key=version_from_branch)
 
     @property
     def username(self):
--- a/.gitignore
+++ b/.gitignore
@@ -90,3 +90,6 @@ ENV/
 
 # CPython checkout for cherry_picker.
 cherry_picker/cpython
+
+# pytest
+.pytest_cache/
--- a/cherry_picker/cherry_picker/test.py
+++ b/cherry_picker/cherry_picker/test.py
@@ -76,6 +76,19 @@ def test_sorted_branch(os_path_exists, config, input_branches, sorted_branches):
     assert cp.sorted_branches == sorted_branches
 
 
+@pytest.mark.parametrize('input_branches', [
+    (['3.1', '2.7', '3.x10', '3.6', '']),
+    (['stable-3.1', 'lts-2.7', '3.10-other', 'smth3.6else', 'invalid']),
+])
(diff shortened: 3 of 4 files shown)


cherry_picker/cherry_picker/cherry_picker.py
```diff
@@ -72,10 +73,16 @@ def upstream(self):
 
     @property
     def sorted_branches(self):
+        def version_from_branch(branch):
+            try:
+                return tuple(map(int, re.match(r'^.*(?P<version>\d+(\.\d+)+).*$', branch).groupdict()['version'].split('.')))
+            except AttributeError as attr_err:
+                raise ValueError(f'Branch {branch} seems to not have a version in its name.') from attr_err
+
         return sorted(
```
webknjaz: @Mariatta optionally we could try/except this as well and fall back to returning branches list in their original order.

cherry_picker/cherry_picker/cherry_picker.py
```diff
@@ -72,10 +73,16 @@ def upstream(self):
 
     @property
     def sorted_branches(self):
+        def version_from_branch(branch):
+            try:
+                return tuple(map(int, re.match(r'^.*(?P<version>\d+(\.\d+)+).*$', branch).groupdict()['version'].split('.')))
```
asvetlov: Maybe reading regex template from config is even better?

webknjaz: I see how it's more flexible. I was thinking about this, but IRL regex is easy to screw up: they'll be exposed to the internals and helpless, and it would require them to actually scan through the code to figure out how it's being used, why and how to fix that. It also looks like mixing up higher-level public API with implementation details, which IMHO should be separated. So I'd be cautious when providing the end-users with such tooling.

What do you think?

asvetlov: Maybe invite a syntax like `stable-*` where the star is *version*?

webknjaz: It seems to be more complex than needed

Mariatta:
Sorry for the delay @webknjaz. I intend to review your PRs sometime this week for sure.

webknjaz:
Thanks!

cherry_picker/cherry_picker/cherry_picker.py
webknjaz: @Mariatta what do you think about falling back to unordered branch list?

webknjaz:
Sure, done!

cherry_picker/cherry_picker/cherry_picker.py
Mariatta: Since there is no need to support unordered branch list yet, I'm thinking to keep things as is and not change the behavior.

Mariatta:
Thanks! I made some changes to readme, but the rest looks great. Thanks for working on this 🌮

webknjaz:
@Mariatta thanks! Please let me know when it's on PYPI.

P.S. If you would like it — I can help with automation of release process of dists from Travis CI to PYPI, triggered by pushing tag, for example. Ping me if you're interested :wink:

Mariatta:
> P.S. If you would like it — I can help with automation of release process of dists from Travis CI to PYPI, triggered by pushing tag, for example. Ping me if you're interested 😉

😲 I'm interested. See related issue https://github.com/python/core-workflow/issues/222 😄 
If you can propose the PR, I'll check it out.

Mariatta:
@webknjaz released to PyPI 🌮

PR #254, opened 2018-05-23
Add support for default branch detection

### LINKED ISSUE #250: [cherry-picker] Misleading assumption about the main branch
Hi,

This is probably a follow-up for #225.

On some projects I work on, the repo's default branch is not `master`, but `devel` or smth else. It would be nice to have it configurable and maybe query this information from github repo metadata as a fallback.

webknjaz:
Fix #250


cherry_picker/cherry_picker/cherry_picker.py
```diff
@@ -51,6 +51,8 @@ def __init__(self, pr_remote, commit_sha1, branches,
         if dry_run:
             click.echo("Dry run requested, listing expected command sequence")
 
+        self.remember_current_branch()
```
asvetlov: The method is called only once, let's inline it.

cherry_picker/cherry_picker/cherry_picker.py
```diff
@@ -104,6 +106,7 @@ def run_cmd(self, cmd):
             return
         output = subprocess.check_output(cmd, stderr=subprocess.STDOUT)
         click.echo(output.decode('utf-8'))
+        return output
```
asvetlov: There the function's result is used?
I don't see any application.

cherry_picker/cherry_picker/cherry_picker.py
webknjaz: I think named things are better readable within the code flow, don't you?

cherry_picker/cherry_picker/cherry_picker.py
webknjaz: Yeah, it was used in the previous commit. Now it's not.

cherry_picker/cherry_picker/cherry_picker.py
asvetlov: Well, I don't care too much.
My the only objection is the call *hides* `_start_branch` initialization by moving attribute assignment from constructor to method.
Also, cherry-picker doesn't use underscores for attribute names.
Let's rename `self._start_branch` for consistency.

webknjaz: Yeah, I've also had this concern and then saw this pattern being used all over the place. I think, I can simplify this.

webknjaz:
@Mariatta

Fair enough! I've changed it as suggested and added doc to readme.

PR #264, opened 2018-06-16
Explicit push location to ignore user git config

abadger:
Users can configure what strategy "git push" uses to determine which
remote branch it should push to.  Cherry-picker doesn't work with all of
the git push strategies but we can make explicit what the remote branch
should be which works around that problem.

In my case, I had this in my .gitconfig (perhaps needed long ago when git used a less intuitive push strategy by default:
```
[push]
        default = tracking
```

That caused git to push to the remote that the branch was branched from rather than trying to push to a branch with the same name as the existing, local one.  Being explicit about both the local and remote branchname should just make the behaviour cherry_picker expects work in both default git configurations and customzed configurations.

### DIFF
--- a/cherry_picker/cherry_picker/cherry_picker.py
+++ b/cherry_picker/cherry_picker/cherry_picker.py
@@ -202,7 +202,7 @@ Co-authored-by: {get_author_info_from_short_sha(self.commit_sha1)}"""
     def push_to_remote(self, base_branch, head_branch, commit_message=""):
         """ git push <origin> <branchname> """
 
-        cmd = ['git', 'push', self.pr_remote, head_branch]
+        cmd = ['git', 'push', self.pr_remote, f'{head_branch}:{head_branch}']
         try:
             self.run_cmd(cmd)
         except subprocess.CalledProcessError:


the-knights-who-say-ni:
Hello, and thanks for your contribution!

I'm a bot set up to make sure that the project can legally accept your contribution by verifying you have signed the [PSF contributor agreement](https://www.python.org/psf/contrib/contrib-form/) (CLA).

Unfortunately we couldn't find an account corresponding to your GitHub username on [bugs.python.org](https://bugs.python.org/) (b.p.o) to verify you have signed the CLA (this might be simply due to a missing "GitHub Name" entry in your b.p.o account settings). This is necessary for legal reasons before we can look at your contribution. Please follow [the steps outlined in the CPython devguide](https://devguide.python.org/pullrequest/#licensing) to rectify this issue.


When your account is ready, please add a comment in this pull request
and a Python core developer will remove the `CLA not signed` label
to make the bot check again.

Thanks again for your contribution, we look forward to reviewing it!

abadger:
Okay, github nick associated with my bugs.python.org account (Contributor Form: Yes on: 2011-03-15.21:00:00) So this should be ready for the bot to check again.

Mariatta:
Thanks! 🌮

abadger:
Thank you :-)

PR #266, opened 2018-06-16
Validate a branch that we parse when running cherry_picker --continue 

abadger:
When running cherry_picker --continue we count on being able to get
certain information from the branch's name which cherry_picker
constructed earlier.  Verify as best we can that the branch name is one
which cherry_picker could have constructed.

Relies on the changes here: #265
(which does the work of one of the validations)

### DIFF
--- a/cherry_picker/cherry_picker/cherry_picker.py
+++ b/cherry_picker/cherry_picker/cherry_picker.py
@@ -74,7 +74,7 @@ class CherryPicker:
 
     @property
     def sorted_branches(self):
-        """Return the branches to cherry-pick to, sorted by version"""
+        """Return the branches to cherry-pick to, sorted by version."""
         return sorted(
             self.branches,
             reverse=True,
@@ -349,12 +349,15 @@ To abort the cherry-pick and cleanup:
             click.echo(f"Current branch ({cherry_pick_branch}) is not a backport branch.  Will not continue. \U0001F61B")
 
     def check_repo(self):
-        # CPython repo has a commit with
-        # SHA=7f777ed95a19224294949e1b4ce56bbffcb1fe9f
-        cmd = ['git', 'log', '-r', self.config['check_sha']]
+        """
+        Check that the repository is for the project we're configured to operate on.
+
+        This function performs the check by making sure that the sha specified in the config
+        is present in the repository that we're operating on.
+        """
         try:
-            subprocess.check_output(cmd, stderr=subprocess.STDOUT)
-        except subprocess.SubprocessError:
+            validate_sha(self.config['check_sha'])
+        except ValueError:
             raise InvalidRepoException()
 
 
@@ -416,6 +419,9 @@ def cherry_pick_cli(dry_run, pr_remote, abort, status, push, config_path,
 def get_base_branch(cherry_pick_branch):
     """
     return '2.7' from 'backport-sha-2.7'
+
+    raises ValueError if the specified branch name is not of a form that
+        cherry_picker would have created
     """
     prefix, sha, base_branch = cherry_pick_branch.split('-', 2)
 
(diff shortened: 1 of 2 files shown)


cherry_picker/cherry_picker/cherry_picker.py
```diff
@@ -422,10 +417,29 @@ def get_base_branch(cherry_pick_branch):
     """
     return '2.7' from 'backport-sha-2.7'
     """
-    prefix, sep, base_branch = cherry_pick_branch.rpartition('-')
+    prefix, sha, base_branch = cherry_pick_branch.split('-', 2)
+
+    if prefix != 'backport':
+        raise ValueError('branch name is not prefixed with "backport-".  Is this a cherry_picker branch?')
+    if not re.match('[0-9a-f]{7,40}', sha):
```
webknjaz: Could also validate the sha against git (like it's done with a commit in config)

abadger: Done.

cherry_picker/cherry_picker/cherry_picker.py
```diff
@@ -74,12 +74,7 @@ def upstream(self):
 
     @property
     def sorted_branches(self):
-        def version_from_branch(branch):
-            try:
-                return tuple(map(int, re.match(r'^.*(?P<version>\d+(\.\d+)+).*$', branch).groupdict()['version'].split('.')))
-            except AttributeError as attr_err:
-                raise ValueError(f'Branch {branch} seems to not have a version in its name.') from attr_err
-
+        """Return the branches to cherry-pick to, sorted by version"""
```
webknjaz: Nitpick: I prefer PEP 257 compatible docstrings. Please add a period at the end of the sentence.

cherry_picker/cherry_picker/cherry_picker.py
```diff
@@ -422,10 +417,37 @@ def get_base_branch(cherry_pick_branch):
...
-    prefix, sep, base_branch = cherry_pick_branch.rpartition('-')
+    prefix, sha, base_branch = cherry_pick_branch.split('-', 2)
+
+    if prefix != 'backport':
+        raise ValueError('branch name is not prefixed with "backport-".  Is this a cherry_picker branch?')
+
+    if not re.match('[0-9a-f]{7,40}', sha):
+        raise ValueError(f'branch name has an invalid sha: {sha}')
+
+    cmd = ['git', 'log', '-r', sha]
```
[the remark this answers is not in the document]
abadger: I thought about this but something struck me as off about doing that and I couldn't quite put my finger on it yesterday.  This morning I realized it's because check_repo() is not well encapsulated.  The sha is contained as an attribute on the CherryPicker object but the repository is coming from the current working directory.  This wasn't immediately obvious because cherry_picker doesn't give you the option to work on a repository that's not the current working directory but when I thought about making the code generic to both places it triggered my sense that something was wrong.

Now that I know that's what struck me as wrong, I'd like to go ahead and move this to its own function in this PR and submit a second PR that allows setting the working directory from which all the git commands are run.  When I do that, though, functions like get_base_branch() will either need to become methods of CherryPicker or I'll have to pass the path to the repository all the way through the stack.  Does that future plan sound okay?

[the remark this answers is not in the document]
webknjaz: Sounds good

webknjaz:
Hi @Mariatta, could you please review this? It seems complete to me.

Mariatta:
There is merge conflict now.

abadger:
Updated.

Mariatta:
Thanks for the patience! Just finally have a little extra time to review this :)

abadger:
Thanks Mariatta!

PR #272, opened 2018-07-02
Add Python 3.7 to Travis CI

webknjaz:


Mariatta:
Should we test for both 3.7 and 3.7-dev?

webknjaz:
Depending on your preferences there's also `3.8-dev`/`nightly`. Anything would work.

webknjaz:
@Mariatta I've rebased this PR. Should be safe to merge now.

webknjaz:
@Mariatta ^^

Mariatta:
Thanks!! 🌮

webknjaz:
@Mariatta could you please release a new version of cherry-picker? Some ppl have to work with a version from Git because of missing fixes.

webknjaz:
Hm.. something's wrong. This job shouldn't appear in build: https://travis-ci.org/python/core-workflow/jobs/418766372

webknjaz:
Oh..
```yaml
if: false
```
So it didn't work. it probably needs something more falsy.

Mariatta:
> @Mariatta could you please release a new version of cherry-picker? Some ppl have to work with a version from Git because of missing fixes.

Ok. I was wondering if I should wait for the fix of #277 first or not?

Mariatta:
hmm.. Maybe I should revert this first?

webknjaz:
No, it's probably because of duplicate key and travis' parser "saw" only first one or so. It's probably from previous PR, not this one

webknjaz:
And it's not critical, because there's another check preventing actual deploy. I'll fix if-clause anyway.

webknjaz:
I've raised the issue with Travis, because it looks like their bug: travis-ci/travis-conditions#13.

Issue #275, opened 2018-07-13
CalledProcessError attempting to add a blurb


jaraco:
I'm attempting to use blurb for the first time, but right off the bat, I'm getting a CalledProcessError:

```
cpython issue-34108 $ rwt blurb -- -m blurb add
Collecting blurb
  Downloading https://files.pythonhosted.org/packages/41/90/cdce96c5f5767e08a87dad88cb2da0b962c692e826d2083d983d1f6f4019/blurb-1.0.7-py3-none-any.whl (62kB)
    100% |████████████████████████████████| 71kB 374kB/s
Installing collected packages: blurb
Successfully installed blurb-1.0.7
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/var/folders/c6/v7hnmq453xb6p2dbz1gqc6rr0000gn/T/rwt-f3yblb6y/blurb.py", line 1664, in <module>
    main()
  File "/var/folders/c6/v7hnmq453xb6p2dbz1gqc6rr0000gn/T/rwt-f3yblb6y/blurb.py", line 1624, in main
    sys.exit(fn(*filtered_args, **kwargs))
  File "/var/folders/c6/v7hnmq453xb6p2dbz1gqc6rr0000gn/T/rwt-f3yblb6y/blurb.py", line 952, in add
    flush_git_add_files()
  File "/var/folders/c6/v7hnmq453xb6p2dbz1gqc6rr0000gn/T/rwt-f3yblb6y/blurb.py", line 1111, in flush_git_add_files
    subprocess.run(["git", "add", "-f", *git_add_files], stdout=subprocess.PIPE, stderr=subprocess.PIPE).check_returncode()
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/subprocess.py", line 409, in check_returncode
    self.stderr)
subprocess.CalledProcessError: Command '['git', 'add', '-f', '/Users/jaraco/Dropbox/code/public/cpython/Misc/NEWS.d/next/Library/2018-07-13-08-44-52.bpo-34108.RjobUC.rst']' returned non-zero exit status 128.
```

jaraco:
[...]

PR #276, opened 2018-07-14
Better reporting in adverse conditions

jaraco:
In #275, I described an issue where a CalledProcessError occurs but no indication is reported, a condition that likely would have been reported if the output hadn't been suppressed.

Although a simple retry worked around the issue, it may happen again in the future, and it would be nice to know what is the cause. This change enables that.

In the third commit, this change also uses the long-form option to git, making it easier for someone troubleshooting an issue like this one unfamiliar with the meaning of `-f` to understand what's happening.

### DIFF
--- a/blurb/blurb.py
+++ b/blurb/blurb.py
@@ -1108,14 +1108,14 @@ Python News
 git_add_files = []
 def flush_git_add_files():
     if git_add_files:
-        subprocess.run(["git", "add", "-f", *git_add_files]).check_returncode()
+        subprocess.run(["git", "add", "--force", *git_add_files]).check_returncode()
         git_add_files.clear()
 
 git_rm_files = []
 def flush_git_rm_files():
     if git_rm_files:
         try:
-            subprocess.run(["git", "rm", "--quiet", "-f", *git_rm_files]).check_returncode()
+            subprocess.run(["git", "rm", "--quiet", "--force", *git_rm_files]).check_returncode()
         except subprocess.CalledProcessError:
             pass
 
--- a/blurb/blurb.py
+++ b/blurb/blurb.py
@@ -1108,14 +1108,14 @@ Python News
 git_add_files = []
 def flush_git_add_files():
     if git_add_files:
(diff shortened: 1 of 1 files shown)


blurb/blurb.py
```diff
@@ -1108,14 +1108,14 @@ def print(*a, sep=" "):
 git_add_files = []
 def flush_git_add_files():
     if git_add_files:
-        subprocess.run(["git", "add", "-f", *git_add_files], stdout=subprocess.PIPE, stderr=subprocess.PIPE).check_returncode()
```
webknjaz: How about introducing a flag (`--verbose` CLI option or config) for controlling this? Dumping output out would probably confuse users, who are not familiar with Git.

blurb/blurb.py
```diff
@@ -1108,14 +1108,14 @@ def print(*a, sep=" "):
...
         git_add_files.clear()
 
 git_rm_files = []
 def flush_git_rm_files():
     if git_rm_files:
         try:
-            subprocess.run(["git", "rm", "-f", *git_rm_files], stdout=subprocess.PIPE, stderr=subprocess.PIPE).check_returncode()
+            subprocess.run(["git", "rm", "--quiet", "--force", *git_rm_files]).check_returncode()
         except subprocess.CalledProcessError:
             pass
```
webknjaz: The code below could've run `git clean -f ...` to wipe out untracked files.

blurb/blurb.py
jaraco: That wouldn’t have helped me. Since the issue seemed to happen just the once, and then not on subsequent runs, I had no opportunity to retry with verbose. The intention here is to avoid swallowing errors.

The behavior on git add is not to output anything so I don’t expect novice users to see anything different in the nominal case.

larryhastings:
I don't know how the permissions work here--do I need to merge this, or are you able to?  (Would you like to hit the magic green button?  Would you prefer I did?  Do you not care?)

jaraco:
I don't care. Happy to proceed.

PR #281, opened 2018-08-21
Hack falsy if-clase in Travis CI config

webknjaz:
Ref travis-ci/travis-conditions#13
Ref https://github.com/python/core-workflow/pull/272#issuecomment-414726306

cc @Mariatta


webknjaz:
Now it should be ok to release a new version :)

PR #282, opened 2018-08-21
cherry-picker 1.2.1 release

Mariatta:


cherry_picker/cherry_picker/__init__.py
```diff
@@ -1,2 +1,2 @@
 """Backport CPython changes from master to maintenance branches."""
-__version__ = '1.2.1.dev1'
+__version__ = '1.2.1'
```
webknjaz: btw what do you think of having scm-based versioning?

Mariatta:
Huh why does this triggers the "blurb" in 3.5 build?

cherry_picker/cherry_picker/__init__.py
Mariatta: What do you mean by scm-based versioning?

Mariatta:
I'm going to ignore the failed build on Python 3.5.

Mariatta:
@webknjaz So I just need to tag this brach with release `cherry-picker-v1.2.1`, and travis will take care of publishing to PyPI, correct?

cherry_picker/cherry_picker/__init__.py
webknjaz: Git tag being the source of truth for the version. Currently, it's manual update of version in Python module followed by tagging the commit. But there are flows, where it only sits in Git, so it would need you to just push a git tag to trigger the whole flow without extra steps.

webknjaz:
Correct

webknjaz:
Blurb was triggered because it's a PR. Tagged commit will have a separate build in CI, where blurb jobs are going to be hidden.

cherry_picker/cherry_picker/__init__.py
Mariatta: Ah ok. This version info is needed for `flit` though.

Mariatta:
I've made this tag: https://github.com/python/core-workflow/releases/tag/cherry-picker-v1.2.1

Mariatta:
Building now: https://travis-ci.org/python/core-workflow/builds/418801217

cherry_picker/cherry_picker/__init__.py
webknjaz: Right, so when using setup.py I use setuptools_scm to make dist have info about version. It should be possible with flit as well.

Mariatta:
And it's live in PyPI! 🌮 Thanks @webknjaz !!

cherry_picker/cherry_picker/__init__.py
webknjaz: Sounds interesting?

Mariatta: Hmm, I think I'll keep the current workflow for now.
Interesting idea though, I'll have to think about it. Thanks!

Mariatta: Ok I'm interested now in deriving the version info automatically somehow, but I'd like to hear more first of what the workflow would look like.

webknjaz: Well, I need to check what's available for flit, but in general it would be triggered just by pushing a tag and during build something figures out the version and puts it into metadata. setuptools_scm, for example, has a mode, where it'll create the python module from version if the project itself needs to use it (however it's also retrievable via pkg_resources part of setuptools).
In case of setuptools_scm, during development it uses last tag plus commit marker for versioning, but that's configurable.

As for prerequisites, you'll likely need to put a commit with changelog in git and tag afterwards.
Some ppl use tools like towncrier or reno to enforce changelog fragments be submitted along with each PR and then can fail tagged builds if those are not converted into common changelog.

I think I'll try submitting some demo PR with implementation once I have some time.

Mariatta: Ok thanks. Before you go ahead with some demo PR, let me try to digest all of the above and decide whether it's worth the trouble. (sounds more complicated that I wanted) 😅 

I definitely want to further automate this somehow.

webknjaz: Well, let's separate stages here:
1. Removing need to track version in files
2. Integrating changelog generator

PR #287, opened 2018-08-26
[WIP] Move cherry-picker to scm versioning via Git tags

webknjaz:


.travis.yml
```diff
@@ -30,7 +30,7 @@ before_install:
     TARGET_PKG: cherry_picker
   install:
   - &cd-to-project pushd "$TARGET_PKG"
-  - flit install
+  - &flit-install flit install --extras dev,test
```
Mariatta: The command is 
`flit install --deps develop`

It will install dev and tests requirements.

webknjaz: I tried this: see ~10 last commits :(

webknjaz: I tried this: see ~10 last commits :(

webknjaz:
@Mariatta for some reason it says "production" in log all the time: https://travis-ci.com/webknjaz/core-workflow/jobs/142033230#L471
No matter what command runs.

webknjaz:
Okay, so here's what's happening: https://github.com/takluyver/flit/blob/fcdb3ee5009624863011570f3a9dd82cb86cacd7/flit/install.py#L263

It does reqs probe by pre-importing the module and then runs subinstaller to install "production" deps

webknjaz:
I've failed to make it work because of the way flit extracts values from evaluated module :(
So it uses `module.__dict__.get('__version__', None)`, which works with already evaluated data, not triggering `__getattr__` stuff.

Mariatta:
Thanks. Sorry it's not working out with flit.
Looking at how setuptools-scm works now, I think I prefer having explicit version info `__version__ = X.Y.Z`, as part of codebase, instead of deriving it from git tag.
Let me rethink of this workflow a bit more.
Thanks.

webknjaz:
Well, setuptools-scm does work well on its own. The problem is with flit not installing dependencies on time (before trying to mess with imports).
Just to be clear: I've had it actually working great with `flit` when `setuptools-scm` was pre-installed before running `flit install`.

hugovk:
Cherry Picker now lives in its own repo at https://github.com/python/cherry-picker

Please re-open there is still relevant.

PR #288, opened 2018-09-10
Add fake f-strings to blurb.  Works with 3.5!

larryhastings:


Mariatta:
How come travis CI is not run here?

larryhastings:
I don't know anything about it!

vstinner:
I told @larryhastings that you get ride of Python 3.5 support if you want. I'm too lazy to check which operating systems don't provide Python 3.6 or newer. We can revert Python 3.5 support if we can too many complains :-)

larryhastings:
Yes, but I felt like rising to the challenge!  This supports something very similar to f-strings, but still runs under 3.5.  (I tested that!)

larryhastings:
I remind both of you that reviewing this PR would get you on the leaderboard for the dev sprints ;-)

Mariatta:
I don't know... this PR is quite evil literally. 

I just want more people to install Python 3.6 :)

blurb/blurb.py
```diff
@@ -1420,11 +1447,11 @@ def flush_version():
...
                 line = "- Issue #9516 and Issue #9516: avoid errors in sysconfig when MACOSX_DEPLOYMENT_TARGET"
             elif line.title().startswith(("- Request #", "- Bug #", "- Patch #", "- Patches #")):
-                # print("FIXING LINE {}: {!r}".format(line_number), line)
+                # print(f("FIXING LINE {line_number}: {line!r}"))
                 line = "- Issue #" + line.partition('#')[2]
-                # print("FIXED LINE {!r}".format(line))
+                # print(f("FIXED LINE {line_number}: {line!r}"))
             # else:
-            #     print("NOT FIXING LINE {}: {!r}".format(line_number, line))
+            #     print(f("NOT FIXING LINE {line_number}: {line!r}"))
```
ericvsmith: I admire your tenacity for even fixing the comments!

blurb/blurb.py
```diff
@@ -1601,10 +1628,10 @@ def main():
         def handle_option(s, dict):
             name = dict.get(s, None)
             if not name:
-                sys.exit('blurb: Unknown option for {}: "{}"'.format(subcommand, s))
+                sys.exit(f('blurb: Unknown option for {subcommand}: "{s}"'))
```
ericvsmith: I would probably use `{s!r}` here instead of `"{s}"`, but I realize it's outside the scope of your direct changes.

larryhastings: You would change the name of the *variable* from *s* to *str?*

ericvsmith: No, I would use `!r` to add the quotes, instead of adding them manually. I realize it might change the type of quotes from double to single, based on what `s` contains.

ericvsmith:
To address @vstinner 's comment: I think this is a reasonable change as an interim to using f-strings in the future. Once this is done, actually switching to f-strings is trivial.

blurb/blurb.py
larryhastings: Oh!  Sorry.  Tiny font made `s!r` look like `str` on my screen.

Mariatta:
Just going to close and re-open the PR, hoping to re-trigger Travis CI.

PR #297, opened 2018-12-04
Use Co-Authored-By instead of Co-authored-by.

serhiy-storchaka:
I don't know whether this matters, but GitHub itself uses `Co-Authored-By` in approved suggestions.


Mariatta:
I think casing doesn't matter. The examples here uses `Co-authored-by`: https://help.github.com/articles/creating-a-commit-with-multiple-authors/ 

I'm inclined to keep things as is.

Mariatta:
Thanks, but since the examples in GitHub help pages use `Co-authored-by:`, I don't feel the need to change what we have now.

PR #302, opened 2019-01-13
Relax click dependency.

asvetlov:
Allow to install cherry_picker with the modern click version

### DIFF
--- a/cherry_picker/cherry_picker/cherry_picker.py
+++ b/cherry_picker/cherry_picker/cherry_picker.py
@@ -381,11 +381,11 @@ CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help'])
               help=("Path to config file, .cherry_picker.toml "
                     "from project root by default"),
               default=None)
-@click.argument('commit_sha1', 'The commit sha1 to be cherry-picked', nargs=1,
-                default = "")
-@click.argument('branches', help='The branches to backport to', nargs=-1)
+@click.argument('commit_sha1', nargs=1, default="")
+@click.argument('branches', nargs=-1)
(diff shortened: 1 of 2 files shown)


Mariatta:
Thanks! Do you need a release cut?

asvetlov:
It would be nice!

Mariatta:
Can you increase/update the version info, and the changelog?
Similar to https://github.com/python/core-workflow/pull/282/files

After that, create a GitHub tag of the release branch, and it will be automatically released to PyPI.

There's a brief write-up here: https://github.com/python/core-workflow/tree/master/cherry_picker#publishing-to-pypi

Mariatta:
v1.2.2 is now on PyPI

cherry_picker/cherry_picker/cherry_picker.py
```diff
@@ -381,11 +381,11 @@ def check_repo(self):
               help=("Path to config file, .cherry_picker.toml "
                     "from project root by default"),
               default=None)
-@click.argument('commit_sha1', 'The commit sha1 to be cherry-picked', nargs=1,
```
webknjaz: Why remove the explanation? How is this related to the click dependency update?

webknjaz: oh, they changed api..

asvetlov: Explicitly forbade meaningless argument actually. The help was never displayed IIRC

webknjaz: Oh, nevermind. I've already figured out that it fails with this arg under newer Click: https://travis-ci.org/python/core-workflow/jobs/491106736#L270

PR #303, opened 2019-01-24
Cherry-picker 1.2.2 release.

Mariatta:


asvetlov:
Looks like you've released the cherry_picker with my patch.
Thank you so much.
I'm on a business trip now.
I used to think about making a release following your instructions but was too busy, unfortunately.

PR #308, opened 2019-02-15
blurb: get BPO# from branch, add commit template - update the PR

matrixise:
Update of the [pull request](https://github.com/python/core-workflow/pull/184) of @tiran 

Co-authored-by: Christian Heimes <christian@python.org>

PR #312, opened 2019-02-28
Add --bpo and --section flags to "blurb add".

larryhastings:


blurb/LICENSE.txt
```diff
@@ -1,4 +1,4 @@
-blurb version 1.0
+blurb version 1.0.8
 Part of the blurb package.
 Copyright 2015-2018 by Larry Hastings
```
Mariatta:
```suggestion
Copyright 2015-2019 by Larry Hastings
```

larryhastings:
Thanks for the suggestion, Mariatta!

blurb/blurb.py
```diff
@@ -93,7 +93,7 @@
 #.. section: IDLE
 #.. section: Tools/Demos
 #.. section: C API
-
+{section}
```
matrixise: why don't you keep the same principle with the `.. bpo: {bpo}`? at this line, you already have a constructed section.
You build the section at that line https://github.com/python/core-workflow/pull/312/files#diff-1bd652c5f766ddebec6cd6dc8f27b3b1R927

Do you make the construction before because you don't want to have a conditional template? is it right?

matrixise:
in my workflow, I could use that 
```fish
blurb add --bpo=(git rev-parse --abbrev-ref HEAD | sed -e 's/bpo-//')
```

so, I am happy with that solution and we can add some alias for the several sections...

`alias blurbadddoc="blurb add --bpo=(git rev-parse --abbrev-ref HEAD | sed -e 's/bpo-//') --section=Documentation"`

blurb/blurb.py
```diff
@@ -913,24 +915,19 @@ def add():
...
+    if bpo:
+        try:
+            int(bpo)
+        except ValueError:
+            error("blurb add --bpo argument " + repr(bpo) + " is not a valid integer!")
 
-    init_tmp_with_template()
+    if section:
+        if section not in sections:
+            error("blurb add --section argument " + repr(section) + " is not a valid section!")
```
hugovk, line 926: How about printing the valid sections when this fails?


```suggestion
            error("blurb add --section argument " + repr(section) + " is not a valid section! Use one of:\n" + "\n".join(sections))
```


For example:

```console
$ python blurb.py add --bpo 123 --section aaa
Error: blurb add --section argument 'aaa' is not a valid section! Use one of:
Security
Core and Builtins
Library
Documentation
Tests
Build
Windows
macOS
IDLE
Tools/Demos
C API
```

hugovk (approved):
Works well! Needs a sync with `master` as blurb 1.0.8 has since been released.

hugovk:
This PR replaces and will close https://github.com/python/core-workflow/pull/184 and https://github.com/python/core-workflow/pull/308.

hugovk:
Here's a rebased branch: https://github.com/python/core-workflow/compare/master...hugovk:bpo-command-line-option2?expand=1

blurb/README.rst
```diff
@@ -247,6 +247,13 @@ part of the cherry-picking process.
 Changelog
 ---------
 
+1.0.8
+~~~~~
+
+- Added the ``--bpo`` and ``--section`` flags to
```
merwok, line 253: Does it make sense to use the bpo abbreviation, or would it be better as `issue` now?

hugovk:
This will need recreating for the new repo at https://github.com/python/blurb.

I've opened https://github.com/python/blurb/issues/6 for this.

PR #311, opened 2019-02-22
cherry-picker: Run Travis CI test on Windows

encukou:
I've set up a Windows build on Travis before, maybe I can help #296 a bit!
There's no Python pre-installed, but there's a package manager, `choco`, that can install it.

Currently, the tests fail because they seem to require Git's `user.email` and `user.name` to be set globally. They fail on my dev machine as well, since I don't have an e-mail set globally (I use different e-mail addresses on different repos).
Sadly, I can't devote time to fixing that issue, but I can help to reproduce or explain more if needed.

To try out changes privately, here's what I do:
* [Enable Travis on your fork](https://travis-ci.org/account/repositories) of the repo
* Uncheck "Build pused branches" and "Build pushed pull requests" on your fork's Travis settings (you'll want to trigger builds manually)
* Push code you want to test to *your fork's master branch*
* Use "Trigger build in Travis's "More options" menu to trigger the build
* Cancel any builds you're not interested in (for example, all the Linux builds) to save both time and processing

PR #315, opened 2019-03-15
Cherry-picker: Remove BACKPORT_COMPLETE. UNSET state after finished.

Mariatta:


Mariatta:
@webknjaz @asvetlov can you take a look. miss-islington is seeing the "BACKPORT_COMPLETE" status, but couldn't continue because it is not an allowed state. I think it should be an allowed state, or did you have different expectation.

webknjaz:
Checking it now.

webknjaz:
@Mariatta I expected that state would be reset if everything is completed successfully. How about calling `reset_state()` after line 391 instead?

webknjaz:
Any non-allowed states are supposed idicate that something went wrong.

Mariatta:
Ok I've removed BACKPORT_COMPLETE. unset will be called instead, and using the UNSET state.

Mariatta:
uugh don't know what's the deal with travis 😟 Tests-passing-locally-for-me™️

webknjaz:
Oh, that's weird... Did you run all tests locally or just related?

Mariatta:
Hmm I ran everything `pytest cherry_picker/test.py -v` in the cherry_picker directory

webknjaz:
I checked this branch locally as well and it passes.

webknjaz:
@Mariatta just to be on the safe side, could you please clear the cache in Travis CI? It's here https://travis-ci.org/python/core-workflow/caches

Mariatta:
I've cleared cache and restart the build, still no luck 🤕 Anyone else can try this locally?

webknjaz:
@Mariatta could you please merge #311 and then rebase this one? The test suite didn't use new fixtures in some tests and it could improve the situation, I think (hope).

webknjaz:
@Mariatta I know what the problem is!

So `git log -r dc896437c8efe5a4a5dfa50218b7a6dc0cbe2598` is executed to check the repo.
But Travis CI only gets 50 last commits by default, not the whole tree. So this commit is now further than 50 commits and the check fails.

webknjaz:
Solutions:
1) Change the sha in config to something newer
2) Find `&install-and-test-cherry-picker` on line 28 in `.travis.yml` and extend the mapping to contain Git config adjustment:
```yaml
- &install-and-test-cherry-picker
  ...
  git:
    depth: 200
  ...
```

cherry_picker/cherry_picker/cherry_picker.py
```diff
@@ -16,19 +16,22 @@
...
-                          "{config[team]}/{config[repo]}/pulls")
-DEFAULT_CONFIG = collections.ChainMap({
-    'team': 'python',
-    'repo': 'cpython',
-    'check_sha': '7f777ed95a19224294949e1b4ce56bbffcb1fe9f',
-    'fix_commit_msg': True,
-    'default_branch': 'master',
-})
+CREATE_PR_URL_TEMPLATE = (
+    "https://api.github.com/repos/" "{config[team]}/{config[repo]}/pulls"
```
webknjaz: I believe, one-line style doesn't require two chunks to be separated
```suggestion
    "https://api.github.com/repos/{config[team]}/{config[repo]}/pulls"
```

cherry_picker/cherry_picker/test.py
```diff
@@ -39,216 +52,241 @@ def changedir(d):
...
+    [
+        (["3.1", "2.7", "3.10", "3.6"], ["3.10", "3.6", "3.1", "2.7"]),
+        (
+            ["stable-3.1", "lts-2.7", "3.10-other", "smth3.6else"],
+            ["3.10-other", "smth3.6else", "stable-3.1", "lts-2.7"],
+        ),
+    ],
+)
+@mock.patch("os.path.exists")
+@mock.patch("cherry_picker.cherry_picker.validate_sha")
```
webknjaz: Why?

Mariatta: These tests aren't testing validate_sha itself, but the other functionalities. I think it is ok to be mocked.

Mariatta:
hm, but changing it to recent sha also only delays the problem, same issue will happen again in the future.

cherry_picker/cherry_picker/test.py
webknjaz: I disagree: more integrated tests is better. Testing a combination of things makes it all more robust. Especially when it costs nothing.
As in

webknjaz:
@Mariatta I think you should just make git clone deeper. Or if you're worried about the delay, it's possible to disable `--depth=` in Travis CI so that it'd clone the whole thing.

Mariatta:
Thanks, I've disabled git depth in travis CI and reverted the mocks.

webknjaz:
@Mariatta great!

P.S. One small feedback about `black out`: as you can see it messes up unrelated stuff making it impossible to track down (by looking at diff) what was the intended change in the PR. It's not completely non-atomic. I'd recommend applying black to the project separately and then when PR is on top of that, using back will only generate related changes :)

Mariatta:
Thanks!

Issue #329, opened 2019-05-30
Should blurb and cherry_picker move to their own repos?


Mariatta:
When we started with blurb and cherry-picker, they were simply command line tools and not published to PyPI.

Now blurb and cherry-picker are in PyPI independently, each have their own chaneglog, their own License (blurb is BSD, cherry-picker is Apache), their own readme, their own owners; perhaps they should be split into their own repos?

One inconvenience I encountered while both are under the same repo, is https://github.com/python/core-workflow/issues/313. I wanted to move cherry-picker's documentation to readthedocs, but readthedocs support one documentation per repo. 

I'm also wondering how having two packages in the same repo affect GitHub's dependency graph 🤔  Perhaps there is no real dependency other than blurb-it depends on blurb, and miss-islington depends on cherry-picker.

asvetlov:
I think the separation makes sense.

brettcannon:
Honestly, I think we should split the projects out and then shut down the repo. This was partially an experiment to see if it would work out better than core-workflow@ and I don't it provides much at this point beyond a place for cherry-picker and blurb. 😄

Mariatta:
Ok, I can look into moving cherry-picker out. Don't know about shutting down repo. I've been telling people to file general workflow-related issues here, the ones not specific to bots.  Just need a different place for that purpose..

brettcannon:
@Mariatta I would say issues should potentially be brought up on discuss.python.org, but we can worry about what to do after cherry-picker and blurb move.

@larryhastings FYI

Mariatta:
I've created a project board for things need doing: https://github.com/python/core-workflow/projects/2

Mariatta:
@larryhastings Can I go ahead and move blurb out? Or you want to do this yourself?

Mariatta:
While I prefer using discourse for discussions over mailing lists, sometimes we need issues to reflect tasks/bugs where we can assign people, and close when the task is done. Discourse doesn't seem to be a natural place for this.

I can't even close my own Discourse post, an admin needs to do it.

So I think this repo is still needed for core-workflow "issues" 🤔

hugovk:
I think we should move blurb to its own repo, and keep this one for issues.

Shall we do this before setting up automated deploys for blurb in this repo? https://github.com/python/core-workflow/issues/530

Mariatta:
Yes I would suggest moving it out first so we don't need to repeat the process later on.

hugovk:
[...]

PR #330, opened 2019-05-30
Remove cherry-picker from core-workflow.

Mariatta:
It has moved to its own repo.
Open issues have been transferred.

Related to #329


.travis.yml
```diff
@@ -88,33 +43,7 @@ jobs:
...
-      PATH: >-
-        /c/Python37:/c/Python37/Scripts:$PATH
-      TARGET_PKG: cherry_picker
-
-  - <<: *deploy-base
-    <<: *run-if-cherry-picker
-    env:
-      TARGET_PKG: cherry_picker
-
-  - <<: *deploy-base
```
webknjaz: @Mariatta I believe that the deploy base shouldn't have been removed from here. Of course, it doesn't affect anything because auto-deployment hasn't been enabled for blurb. But if anybody will want to use it, they'd have to recover that base mapping.

PR #403, opened 2021-05-12
blurb: don't overwrite output when not necessary

dmerejkowsky:
This speeds up the build of the Python documentation
quite a bit, since the NEWS file is huge and causes
sphinx to rebuild its doctree each time the doc
is built, regardless of the other changes

### DIFF
--- a/blurb/blurb.py
+++ b/blurb/blurb.py
@@ -49,10 +49,12 @@ import builtins
 from collections import OrderedDict
 import glob
 import hashlib
+import io
 import inspect
 import itertools
 import math
 import os
+from pathlib import Path
 import re
 import shlex
 import shutil
@@ -1068,9 +1070,11 @@ To force it to overwrite, use -f.
 
 
 def write_news(output, *, versions):
+    buff = io.StringIO()
+
     def print(*a, sep=" "):
         s = sep.join(str(x) for x in a)
-        return builtins.print(s, file=news)
+        return builtins.print(s, file=buff)
 
     print ("""
 +++++++++++
@@ -1131,7 +1135,20 @@ Python News
             print(text)
     print()
     print("**(For information about older versions, consult the HISTORY file.)**")
-    news.close()
+
+
+    new_contents = buff.getvalue()
+
+    # Only write in `output` if the contents are different
+    # This speeds up subsequent Sphinx builds
+    try:
+        previous_contents = Path(output).read_text(encoding="UTF-8")
+    except (FileNotFoundError, UnicodeError):
+        previous_contents = None
+    if new_contents != previous_contents:
+        Path(output).write_text(new_contents, encoding="UTF-8")
+    else:
+        builtins.print(output, "is already up to date")
 
 
 git_add_files = []
--- a/blurb/blurb.py
+++ b/blurb/blurb.py
@@ -1064,8 +1064,10 @@ To force it to overwrite, use -f.
         builtins.print("You already have a", repr(output), "file.")
         require_ok("Type ok to overwrite")
 
-    news = open(output, "wt", encoding="utf-8")
+    write_news(output, versions=versions)
 
+
+def write_news(output, *, versions):
     def print(*a, sep=" "):
         s = sep.join(str(x) for x in a)
         return builtins.print(s, file=news)


dmerejkowsky:
Poke @JulienPalard

dmerejkowsky:
Example of speed up gained from the patch

Without this patch, building the French translation after changing one `.po` takes 30 seconds

With this patch, it only takes 12.

JulienPalard:
How did you tested this?

I'm testing it using:

    $ hyperfine '../../core-workflow/blurb/blurb.py merge -f build/NEWS'

and I'm getting similar results with master vs your branch:

  Time (mean ± σ):      2.767 s ±  0.300 s    [User: 2.743 s, System: 0.021 s]

vs

  Time (mean ± σ):      2.683 s ±  0.288 s    [User: 2.662 s, System: 0.020 s]

JulienPalard:
(You can rebase on top of master if you want the CI to have a chance to pass)

encukou:
> I'm getting similar results with master vs your branch

The speedup comes later. With this change, blurb won't update the file unnecessarily, which means *Sphinx* will then use a cache instead of parsing it again.

dmerejkowsky:
Thanks @encukou, that's a perfect description of what I was trying to achieve :)

dmerejkowsky:
Rebased but I need an approval for the workflow to run. (BTW, thanks for the CI fix)

encukou (approved):
The changes look good and I get the same output as with the current blurb.
This does load the entire `NEWS` file into memory twice, but I don't think that's an issue nowadays.

JulienPalard:
> The speedup comes later. With this change, blurb won't update the file unnecessarily,

So why `hyperfine` did not catched up the speedup? As hyperfine runned blurb multiple times, if succequent runs are faster, I should have spoted the difference?

Also got this today while trying:

```
$ ../../core-workflow/blurb/blurb.py merge build/NEWS
Traceback (most recent call last):
  File "/home/mdk/clones/python/cpython/Doc/../../core-workflow/blurb/blurb.py", line 1705, in <module>
    main()
  File "/home/mdk/clones/python/cpython/Doc/../../core-workflow/blurb/blurb.py", line 1665, in main
    sys.exit(fn(*filtered_args, **kwargs))
  File "/home/mdk/clones/python/cpython/Doc/../../core-workflow/blurb/blurb.py", line 1068, in merge
    write_news(output, versions=versions)
  File "/home/mdk/clones/python/cpython/Doc/../../core-workflow/blurb/blurb.py", line 1138, in write_news
    with open(output, "r", encoding="utf-8") as file:
FileNotFoundError: [Errno 2] No such file or directory: '/home/mdk/clones/python/cpython/Doc/build/NEWS'
```

dmerejkowsky:
> So why hyperfine did not catched up the speedup

No clue. Measuring performance is hard ... But as stated above, the goal is to speed up the sphinx build when blurb is called, not blurb itself.

> Also got this today while trying [...]

I used this 

```
cd cpython/Doc
make BLURB=/mnt/data/dmerej/src/python-core-workflow/blurb/blurb.py  html
```

JulienPalard:
I'm testing with `blurb -f merge` because that's what the doc's Makefile is using.

JulienPalard:
> But as stated above, the goal is to speed up the sphinx build when blurb is called, not blurb itself.

Oh you mean to avoid cache thinking the file has changed by not touching it? Got it now \o/ I had the feeling I was not understanding something ... I was not understanding something!

Can you please just double check the exception on `merge`? I don't want the next docs build to fail :p

dmerejkowsky:
> Can you please just double check the exception on merge? I don't want the next docs build to fail :p

Me neither - but I can't reproduce the exception. This is weird. If you're afraid of breakage, we could push a beta version of blurb on pypi and check nothing breaks, no ?

JulienPalard:
I'd look at it on my machine, as I can reproduce it, but not right now, I'm attending the trainers summit :D

JulienPalard:
OK my reproducer is:

    rm -f build/NEWS; ../../core-workflow/blurb/blurb.py merge -f build/NEWS

which make sense: if the file does not exists, you can't compare it. what about:

```python
new_contents = buff.getvalue()
try:
    previous_contents = Path(output).read_text(encoding="UTF-8")
except (FileNotFoundError, UnicodeError):
    previous_contents = None
if new_contents != previous_contents:
    Path(output).write_text(new_contents, encoding="UTF-8")
else:
    builtins.print(output, "is already up to date")
```

dmerejkowsky:
Yup, that sounds better :)

JulienPalard (approved):

JulienPalard:
Thanks a lot @dmerejkowsky!

PR #451, opened 2022-05-08
Replace `gh-issue-NNNN:` with `gh-NNNN:` in the output

ezio-melotti:
I noticed in python/cpython#92448 that the output shows `gh-issues-NNNN:` instead of `gh-NNNN:`.  I had initially proposed to use `gh-issue` instead of just `issue` in the content of the entry because in that case is ambiguous and people might have added the bpo number.  In the output there is no such problem, and `gh-NNNN` is enough.

AFAIU if this PR is applied, it will update older news entries too, right?

### DIFF
--- a/blurb/blurb.py
+++ b/blurb/blurb.py
@@ -1113,7 +1113,7 @@ Python News
             if metadata.get("gh-issue"):
                 issue_number = metadata['gh-issue']
                 if int(issue_number):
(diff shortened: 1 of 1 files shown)


erlend-aasland (approved):
Makes sense. Does this also need a doc update?

ezio-melotti:
Which docs do you mean?

Issue #450, opened 2022-05-06
`python/cpython` labels grouping


ezio-melotti:
There are two ways to group labels on GitHub: by color and by using a common prefix.  Even though the color-coding helps, without a common prefix, labels that belong in the same category are not grouped together in the "Labels" dropdown in the sidebar.  On the other hand, adding a prefix to all labels makes them longer and adds clutter.

On `python/cpython`, we currently have [61 labels](https://github.com/python/cpython/labels), that include misc labels (like `easy`, `pending`, etc.) and a few main groups, including:
* types (uses the `type-*` prefix)
* experts (uses the `expert-*` prefix)
* OS (uses the `OS-*` prefix)
* version number (no prefix/grouping needed -- they all start with 3)
* "dir" (no prefix)

The "dir" category currently has 5 labels: `stdlib`, `docs`, `tests`, `interpreter-core`, and `extension-modules`, and they indicate what part of the source tree is affected, however since they have no prefix, they are not grouped.  I'm not sure if `dir-*` would be a good prefix, but I can't think of anything better.

Another idea suggested on Discord is to use emojis as prefixes for some categories, since they only take 1 character, are easily recognizable, and can convey some semantic meaning as well.

Some suggestions:
* :open_file_folder: can be used as a prefix for the "dir" group
* :snake: can be used as a prefix for experts
* if we want to replace the `type-*` prefix:
  * :orange_book: for `crash` and `security`, :blue_book: for `bugs`, and :green_book: for `feature requests`
  * :red_circle: :orange_circle: :large_blue_circle: :green_circle: / :red_square: :orange_square: :blue_square: :green_square:
  * :file_cabinet: or :card_file_box: if we want the same emoji for all items
* if we want to replace the `OS-*` prefix: :computer: or :desktop_computer: 
* we could keep some text-based prefix, and not all labels need a prefix
* we could add prefixes to create some mini-groups:
  * ❌ for `invalid`/`spam`
  * :exclamation: for `release blocker`/`deferred blocker`
  * :robot: for labels that trigger bot actions
* some bots/tools might need to be updated if we rename labels

The main goals I want to accomplish are:
* group the "dir" labels
* make related labels easier to find in the dropdown
* make label names shorter
* make the categories clearer

[...]

PR #455, opened 2022-05-22
Add an issue template to request label changes.

ezio-melotti:
Since there are often proposal about adding/removing/updating labels on `python/cpython`, I created this template in order to formalize the process a bit.  I also added a `labels` label to track this type of issues.


erlend-aasland (approved):

erlend-aasland:
We also need to mention this in the devguide.

ezio-melotti:
Once things settle a bit I'll update the devguide too.  There are still a few open issues, including:
* https://github.com/python/core-workflow/issues/450
* https://github.com/python/devguide/issues/821

PR #463, opened 2022-06-21
Mention org owners in the new triager template

encukou:
The devguide [suggests](https://github.com/python/devguide/pull/889/files) that owners should be mentioned when adding new members, so it would make sense to add the mention to the template.

@python/organization-owners, is it a good idea, or are you getting notifications some other way?


ewdurbin (approved):
This makes sense!

PR #473, opened 2022-07-24
Loosen README check for CPython forks

arhadthedev:
This PR supersedes gh-291. A quote from https://github.com/python/core-workflow/pull/291#issuecomment-423268895 for context:

> If we wanted to accommodate CPython forks, we should not 'bless' Tauthon only. /`Py`/`.`/ would potentially also accommodate things like 'Brython'. But it is not clear to me that we want to do this, as it would subvert part of the apparent purpose of the function that uses the RE.

@terryjreedy I believe that `This is Python version 2.7 as modified and augmented by the xxx fork thereof` wording is long and clumsy, and so perceived as *This is Python something something... Looks like a wrong repo*.

### DIFF
--- a/blurb/blurb.py
+++ b/blurb/blurb.py
@@ -701,7 +701,7 @@ def run(s):
     return process.stdout.decode('ascii')
 
 
-readme_re = re.compile(r"This is Python version [23]\.\d").match
+readme_re = re.compile(r"This is \w+thon version [23]\.\d").match
 
 def chdir_to_repo_root():
     global root
--- a/blurb/blurb.py
+++ b/blurb/blurb.py
@@ -701,7 +701,7 @@ def run(s):
     return process.stdout.decode('ascii')
 
 
-readme_re = re.compile(r"This is \w+ version \d\.\d+").match
+readme_re = re.compile(r"This is \w+ version \d+\.\d+").match
 
 def chdir_to_repo_root():
     global root
--- a/blurb/blurb.py
+++ b/blurb/blurb.py
@@ -701,7 +701,7 @@ def run(s):
     return process.stdout.decode('ascii')
 
 
-readme_re = re.compile(r"This is \w+thon version [23]\.\d").match
+readme_re = re.compile(r"This is \w+ version \d\.\d+").match
 
 def chdir_to_repo_root():
     global root


terryjreedy:
The most important part of my previous comment, four years ago, that review was needed from Larry Hastings (or someone who understood the current code) ,was ignored.  Until it is determined whether *any* change to the re is acceptable, it is useless to discuss the exact change.  This PR seems premature.  I think I should have suggested a core-workflow issue.

arhadthedev:
According to dates in `user:larryhastings author:larryhastings` search query [for commits](https://github.com/search?o=desc&q=user%3Alarryhastings+author%3Alarryhastings&s=author-date&type=Commits) and [for issues](https://github.com/search?o=desc&q=user%3Alarryhastings+author%3Alarryhastings&s=updated&type=Issues), he fortunately seems to be relatively active, just dedicated exclusively to other projects.

arhadthedev:
@larryhastings Would you mind to review and possibly merge this PR or gh-291 alternative please? Both are one-liners.

arhadthedev:
@larryhastings ping?

CAM-Gerlach:
FWIW, the tauthon repo hasn't seen any PR or commit activity in nearly 6 months, and only a handful of commits all by a single person in the past year+ since the final release of CPython 2.7, so it's unclear how useful this will be to them at this point. However, it could be useful for other forks that have names ending in `thon` (though many don't, e.g. Pyston, no-gil, etc.)

I've pinged @larryhastings through other channels, but I also want to mention @gpshead who previously reviewed and approved a previous fairly small PR similar to this one, given this is seemingly just a trivial regex.

blurb/blurb.py
```diff
@@ -701,7 +701,7 @@ def run(s):
     return process.stdout.decode('ascii')
 
 
-readme_re = re.compile(r"This is Python version [23]\.\d").match
+readme_re = re.compile(r"This is \w+thon version [23]\.\d").match
```
ezio-melotti, line 704:
```suggestion
readme_re = re.compile(r"This is \w+ version \d\.\d+").match
```
What about relaxing it a bit more to prevent bikeshedding on the exact names to be matched?
This should still fulfill its intended purpose for CPython and might allow other forks to use blurb, with the understanding that it remains CPython-specific and there are no explicit guarantees that it will work with other repos.

ericvsmith: I agree. Do we really care what name is used here? For anyone arguing for strictness: what error is this trying to prevent?

arhadthedev:
@larryhastings ping?

hugovk (approved):

blurb/blurb.py
```diff
@@ -701,7 +701,7 @@ def run(s):
     return process.stdout.decode('ascii')
 
 
-readme_re = re.compile(r"This is Python version [23]\.\d").match
+readme_re = re.compile(r"This is \w+ version \d\.\d+").match
```
ericvsmith, line 704: I think this change is good. I'm not sure we need to restrict the version to one digit in the major version, but we can always change that later if it becomes an actual problem.

ericvsmith (approved):

blurb/blurb.py
CAM-Gerlach:
```suggestion
readme_re = re.compile(r"This is \w+ version \d+\.\d+").match
```

If we don't want to restrict the major version to one digit

CAM-Gerlach (approved):

arhadthedev:
ping

Issue #504, opened 2023-05-19
Blurb: validate GitHub issue number


hugovk:
# The short story

It would be nice if we checked GitHub issue numbers are valid.

# Long version

Thanks for blurb!

If you accidentally use an (old) GitHub _PR_ number instead of the GitHub _issue_ number, the CI docs build will fail:

```
Building NEWS from Misc/NEWS.d with blurb
PATH=./venv/bin:$PATH sphinx-build -b html -d build/doctrees  -j auto -q -W --keep-going . build/html 
Error: ../build/NEWS:7[15](https://github.com/python/cpython/actions/runs/5021764661/jobs/9004477237#step:7:16): ERROR: The GitHub ID '13[17](https://github.com/python/cpython/actions/runs/5021764661/jobs/9004477237#step:7:18)2' seems too low -- use :issue:`...` for BPO IDs
make: *** [Makefile:50: build] Error 1
```

Re:
* https://github.com/python/cpython/actions/runs/5021764661/jobs/9004477237
* https://github.com/python/cpython/pull/13172#issuecomment-1554177530

This is validated in https://github.com/python/cpython/blob/616fcad6e2e10b0d0252e7f3688e61c468c54e6e/Doc/tools/extensions/pyspecific.py#L69-L79

Shall we add some validation to Blurb, to fail sooner?

We could also allow old BPO issue numbers similar to https://github.com/python/cpython/blob/616fcad6e2e10b0d0252e7f3688e61c468c54e6e/Doc/tools/extensions/pyspecific.py#L54-L63, but for new NEWS files created today, should we require only the corresponding GitHub issue number?

terryjreedy:
Blurb twice says it wants a gh # and an issue #.  Would be nice if it could enforce both, at least when online.
```
# Please enter the relevant GitHub issue number here:
#
.. gh-issue: 
```
'here' should be changed to 'below'.

menkotoglou:
I can work on this.

menkotoglou:
Getting back to this @hugovk, after some research seems like we can follow two different directions here.

The validation [here](https://github.com/python/cpython/blob/616fcad6e2e10b0d0252e7f3688e61c468c54e6e/Doc/tools/extensions/pyspecific.py#L69-L79), doesn't really resolve the case that the issue doesn't exist. This only checks if the issue, is less than the MINIMUM_ISSUE_NUMBER.

We could the following:

[...]

PR #519, opened 2023-12-17
Accept GitHub issues numbered only 32426 or above

menkotoglou:
Fixes https://github.com/python/core-workflow/issues/504.

Iterates through the template file to find the gh-issue line and checks number if it's in the accepted issue number range.

This is how an unsuccessful attempt looks like:

```sh
❯ /bin/python3 /path/core-workflow/blurb/blurb.py

Error: The gh-issue number should be 32426 or above.

[Hit return to retry (or Ctrl-C to abort)>
```

### DIFF
--- a/blurb/blurb.py
+++ b/blurb/blurb.py
@@ -489,7 +489,7 @@ class Blurbs(list):
                 # we see in the blurb file, which is a
                 # better user experience.
                 if key == "gh-issue" and int(value) < lowest_possible_gh_issue_number:
-                    throw("The gh-issue number must be 32426 or above, not a PR number.")
+                    throw(f"The gh-issue number must be {lowest_possible_gh_issue_number} or above, not a PR number.")
 
                 if key in issue_keys:
                     try:
--- a/blurb/blurb.py
+++ b/blurb/blurb.py
@@ -956,13 +956,14 @@ Add a blurb (a Misc/NEWS.d/next entry) to the current CPython repo.
             failure = str(e)
 
         if not failure:
-            with open(tmp_path, "rt", encoding="utf-8") as file:
-                for line in file:
-                    if line.startswith(".. gh-issue:"):
-                        issue_number = line.split(":")[1].strip()
-                        if issue_number.isdigit() and int(issue_number) < 32426:
-                            failure = "The gh-issue number should be 32426 or above."
-                            break
+            assert len(blurb)  # if parse_blurb succeeds, we should always have a body
+            if len(blurb) > 1:
+                failure = "Too many entries!  Don't specify '..' on a line by itself."
+            gh_issue_number = blurb[0][0]["gh-issue"]
+            if not gh_issue_number.isdigit():
+                failure = "The gh-issue number must be a number."
+            if int(gh_issue_number) < 32426:
+                failure = "The gh-issue number should be 32426 or above."
 
         if failure:
             print()
--- a/blurb/blurb.py
+++ b/blurb/blurb.py
@@ -486,6 +486,9 @@ class Blurbs(list):
                 # we'll complain about the *first* error
                 # we see in the blurb file, which is a
                 # better user experience.
+                if key == "gh-issue" and int(value) < 32426:
+                    throw("The gh-issue number should be 32426 or above.")
+
                 if key in issue_keys:
                     try:
                         int(value)
@@ -959,11 +962,6 @@ Add a blurb (a Misc/NEWS.d/next entry) to the current CPython repo.
             assert len(blurb)  # if parse_blurb succeeds, we should always have a body
             if len(blurb) > 1:
                 failure = "Too many entries!  Don't specify '..' on a line by itself."
-            gh_issue_number = blurb[0][0]["gh-issue"]
-            if not gh_issue_number.isdigit():
-                failure = "The gh-issue number must be a number."
-            if int(gh_issue_number) < 32426:
-                failure = "The gh-issue number should be 32426 or above."
 
         if failure:
             print()
--- a/blurb/blurb.py
+++ b/blurb/blurb.py
@@ -489,7 +489,7 @@ class Blurbs(list):
                 # we see in the blurb file, which is a
                 # better user experience.
                 if key == "gh-issue" and int(value) < lowest_possible_gh_issue_number:
-                    throw("The gh-issue number should be 32426 or above.")
+                    throw("The gh-issue number must be 32426 or above, not a PR number.")
 
                 if key in issue_keys:
                     try:
--- a/blurb/blurb.py
+++ b/blurb/blurb.py
@@ -473,6 +473,8 @@ class Blurbs(list):
 
             no_changes = metadata.get('no changes')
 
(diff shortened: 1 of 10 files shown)


blurb/blurb.py
```diff
@@ -956,9 +956,13 @@ def init_tmp_with_template():
             failure = str(e)
 
         if not failure:
-            assert len(blurb) # if parse_blurb succeeds, we should always have a body
-            if len(blurb) > 1:
-                failure = "Too many entries!  Don't specify '..' on a line by itself."
```
ezio-melotti, line 961: Should these checks be kept too?

blurb/blurb.py
```diff
@@ -956,9 +956,13 @@ def init_tmp_with_template():
...
+            with open(tmp_path, "rt", encoding="utf-8") as file:
```
ezio-melotti, line 959: The file seems to be already read in `blurb.load/parse`, so you should be able to access it (or even the issue number directly) directly from the instance without opening and parsing the file again.

blurb/blurb.py
```diff
@@ -956,9 +956,13 @@ def init_tmp_with_template():
...
+                for line in file:
+                    if line.startswith(".. gh-issue:"):
+                        issue_number = line.split(":")[1].strip()
+                        if issue_number.isdigit() and int(issue_number) < 32426:
```
ezio-melotti, line 963: Should this also handle (and reject) non-numeric issue numbers?

menkotoglou: Hey @ezio-melotti, that's true. I'll split the two if statements to show different errors.

blurb/blurb.py
```diff
@@ -956,9 +956,14 @@ def init_tmp_with_template():
...
         if not failure:
-            assert len(blurb) # if parse_blurb succeeds, we should always have a body
+            assert len(blurb)  # if parse_blurb succeeds, we should always have a body
             if len(blurb) > 1:
                 failure = "Too many entries!  Don't specify '..' on a line by itself."
+            gh_issue_number = blurb[0][0]["gh-issue"]
+            if not gh_issue_number.isdigit():
+                failure = "The gh-issue number must be a number."
+            if int(gh_issue_number) < 32426:
+                failure = "The gh-issue number should be 32426 or above."
```
hugovk, line 966: Bit of whitespace for readability, and wording:
```suggestion
            gh_issue_number = blurb[0][0]["gh-issue"]

            if not gh_issue_number.isdigit():
                failure = "The gh-issue number must be a number."

            if int(gh_issue_number) < 32426:
                failure = "The gh-issue number must be 32426 or above."
```

hugovk (commented):
Please could you add some tests?

Hopefully we can get https://github.com/python/core-workflow/pull/520 merged soon, then we can add unit tests.

In the meantime, tests are done by adding valid blurb files in the `pass` dir and invalid ones in the `fail` dir at https://github.com/python/core-workflow/tree/main/blurb/tests, and then running `blurb test`.

blurb/blurb.py
menkotoglou: Removed completely from here since we should be raising an exception instead of displaying the failure.

menkotoglou:
Seems like the `not_number` case is already covered in the code. That's validated in the pre-existing `invalid-gh-number.rst` test case. 

Also, the tests wouldn't pass if we didn't raise an exception at the `class Blurbs` level. Decided to move to the class definition level as I'm aligned with this approach.

Finally, retrospectively fixed some old tests that indeed were using smaller GH issue numbers.

blurb/blurb.py
```diff
@@ -486,6 +486,9 @@ def finish_entry():
                 # we'll complain about the *first* error
                 # we see in the blurb file, which is a
                 # better user experience.
+                if key == "gh-issue" and int(value) < 32426:
+                    throw("The gh-issue number should be 32426 or above.")
```
hugovk, line 490: Maybe something like this? It gives a hint they may be using the PR number instead of issue number.

```suggestion
                    throw("The gh-issue number must be 32426 or above, not a PR number.")
```

hugovk (commented):
Thanks! It's nice when adding tests reveals a problem in the original implementation :)

Looks good, here's a wording suggestion.

menkotoglou:
Thanks a lot for the review @hugovk. Suggestion commited!

hugovk (approved):
Thanks!

blurb/blurb.py
```diff
@@ -956,7 +959,7 @@ def init_tmp_with_template():
             failure = str(e)
 
         if not failure:
-            assert len(blurb) # if parse_blurb succeeds, we should always have a body
+            assert len(blurb)  # if parse_blurb succeeds, we should always have a body
```
pradyunsg, line 962: nit: unrelated whitespace change.

blurb/blurb.py
```diff
@@ -486,6 +486,9 @@ def finish_entry():
                 # we'll complain about the *first* error
                 # we see in the blurb file, which is a
                 # better user experience.
+                if key == "gh-issue" and int(value) < 32426:
```
pradyunsg, line 489: Let's put the magic number in a variable with a clear name?

Something like: `lowest_possible_issue_number`? I don't feel strongly about the bikeshed color, as long as the name conveys something about why we're using this specific magic number.

pradyunsg (commented):
A couple of non-blocking suggestions, with one of them being a nitpick that you're welcome to ignore if it's too much effort and a request to name a [magic number](https://en.m.wikipedia.org/wiki/Magic_number_(programming)).

menkotoglou:
Thanks a lot for the review @pradyunsg. Feedback addressed already!

blurb/blurb.py
```diff
@@ -486,6 +488,9 @@ def finish_entry():
                 # we'll complain about the *first* error
                 # we see in the blurb file, which is a
                 # better user experience.
+                if key == "gh-issue" and int(value) < lowest_possible_gh_issue_number:
+                    throw("The gh-issue number should be 32426 or above.")
```
hugovk, line 492: The force push reverted my last suggestion, so here it is again :)
```suggestion
                    throw("The gh-issue number must be 32426 or above, not a PR number.")
```

Tip: don't force push, we squash merge at the end anyway. 
(https://devguide.python.org/getting-started/pull-request-lifecycle/#quick-guide)

blurb/blurb.py
```diff
@@ -486,6 +488,9 @@ def finish_entry():
                 # we'll complain about the *first* error
                 # we see in the blurb file, which is a
                 # better user experience.
+                if key == "gh-issue" and int(value) < lowest_possible_gh_issue_number:
+                    throw("The gh-issue number must be 32426 or above, not a PR number.")
```
pradyunsg, line 492:
```suggestion
                    throw(f"The gh-issue number must be {lowest_possible_gh_issue_number} or above, not a PR number.")
```

menkotoglou:
Hey @pradyunsg, @hugovk, you think we're OK to merge this?

hugovk:
Yes, let's merge. Thank you for your contribution!

PR #520, opened 2023-12-18
blurb: Add tests and run on CI

hugovk:
This splits out the testing added in https://github.com/python/core-workflow/pull/499 to its own PR, to make reviewing of both easier.

These tests are useful even if #499 is not merged.

### DIFF
--- a/blurb/tests/test_blurb.py
+++ b/blurb/tests/test_blurb.py
@@ -5,29 +5,49 @@ import blurb
 
 
 @pytest.mark.parametrize(
-    "section, expected",
+    "section",
     (
-        ("C API", "C API"),
-        ("Core and Builtins", "Core and Builtins"),
-        ("Library", "Library"),
-        ("Tools/Demos", "Tools-Demos"),
+        "C API",
+        "Core and Builtins",
+        "Library",
     ),
 )
-def test_sanitize_section(section: str, expected: str) -> None:
+def test_sanitize_section_no_change(section: str) -> None:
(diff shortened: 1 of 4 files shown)


blurb/tests/test_blurb.py
```diff
@@ -0,0 +1,127 @@
...
+
+@pytest.mark.parametrize(
+    "section, expected",
+    (
+        ("C API", "C API"),
+        ("Core and Builtins", "Core and Builtins"),
+        ("Library", "Library"),
+        ("Tools-Demos", "Tools/Demos"),
+    ),
+)
```
ezio-melotti, line 29: Instead of (almost) duplicating this, what about creating a list of idempotent names and a mapping (or two) of names that get transformed and then test those separately?

I'm also curious to see a few more examples (e.g. `"C api"`, `"Tools\Demos"`, etc.), but those could be added in a separate PR.

ezio-melotti: This still doesn't solve the duplication between `test_unsanitize_section_no_change` and `test_sanitize_section_no_change` though.

hugovk: Updated! https://github.com/python/core-workflow/pull/520/commits/70f600460e435f4c365d2b9d7a91cf527974f82b

hugovk:
Updated following merge of https://github.com/python/core-workflow/pull/519 to replace the arbitrarily chosen (but invalid) issue number 11111 with a valid 55555.

JelleZijlstra (approved):

PR #521, opened 2023-12-18
blurb: Drop support for EOL Python 3.7

hugovk:
Re: https://devguide.python.org/versions/

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -10,8 +10,8 @@ jobs:
     name: ${{ matrix.python-version }}
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
-      - uses: actions/setup-python@v4
+      - uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -6,7 +6,7 @@ jobs:
   build_ubuntu:
     strategy:
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12-dev"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12-dev"]
```
ezio-melotti, line 9: Was the `-dev` suffix still necessary, or is `3.12` enough nowadays?

(Can be addressed in a separate PR if needed.)

ezio-melotti (approved):

.github/workflows/tests.yml
hugovk: Yeah, let's remove it.

